PR #126, opened 2021-09-09
feat(dark-mode): implement automatic dark mode

eddybrando:
The dark theme is activated if the user's system preferred schema is set to **dark**.

The images in the "main content section" (the middle section in desktop) are automatically transformed so that they match the dark theme. Other images are switched to a custom made "dark mode" version.

Closes https://github.com/Qiskit/qiskit.org/issues/2057

---

Related to https://github.com/Qiskit/qiskit.org/issues/2057


eddybrando:
@vabarbosa, I believe that the color of the **reminder block** is consistent design. Here is how it looks in the light theme:

Screenshot 2021-09-20 at 15 45 37

vabarbosa:
> @vabarbosa, I believe that the color of the **reminder block** is consistent design. Here is how it looks in the light theme:
> 

@eddybrando this is what i see in the default/light theme:

image

JRussellHuffman:
I see the same as Va, in terms of the light theme and agree that the reminder block should be updated in the dark theme.
Screen Shot 2021-09-20 at 9 07 08 AM
.

eddybrando:
@vabarbosa, @JRussellHuffman, thank you very much for your reviews!

I made adjustments to every point except to the progress indicator checkmark, since the checkmark part is actually an empty part of the SVG, so its color will be the background's.

I wanted to see if I could change the SVG but I cannot find where this in implemented. @vabarbosa, do you happen to know if this is something we did or if this is coming from somewhere else?

vabarbosa:
> > @eddybrando Can I provide a new SVG? The one currently used is the mathigon default, but we could switch to the carbon default, which has a fill.
> 
> I think we cannot change it. It just comes as-is from Mathigon. Is that assumption right, @vabarbosa?

@eddybrando @JRussellHuffman the progress status a Mathigon component. however, we could replace with our own progress component (which uses our own SVG)

eddybrando:
> > > @eddybrando Can I provide a new SVG? The one currently used is the mathigon default, but we could switch to the carbon default, which has a fill.
> > 
> > 
> > I think we cannot change it. It just comes as-is from Mathigon. Is that assumption right, @vabarbosa?
> 
> @eddybrando @JRussellHuffman the progress status a Mathigon component. however, we could replace with our own progress component (which uses our own SVG)

Although the SVG itself is pretty small (in size), I think this would entail quite a significant effort. There is some logic and animations that we would need to also take into account.

In general, I think we could do that, but probably we should do so as part of a new issue. Also so that this doesn't get blocked from merging. Unless we argue that it's indeed a blocker.

JRussellHuffman:
It's not worth blocking (in my opinion). Maybe we can revisit whenever we have a refactor sprint or a decoration sprint?

vabarbosa (approved):
@eddybrando thank you!
i agree, we can move the progress indicator update to a new issue.

vabarbosa:
this work was reverted from the `main` branch and placed into `feature/dark-mode` branch for merging once switcher is available - https://github.com/qiskit-community/platypus/pull/143

PR #151, opened 2021-09-27
Override mathigon link styling

frankharkins:
Made a mistake in last PR so opening another. Hopefully should fix Qiskit/qiskit.org/issues/2158


frontend/scss/vendor/mathigon.scss
```diff
@@ -403,3 +386,29 @@ x-bio, x-gloss {
...
+    &:hover, &:focus {
+      color: $link-color-tertiary;
+      text-decoration: underline;
+      background: initial;
+      box-shadow: none;
+    }
+
+    &::after {
+      background: initial;
+      background-image: url("data:image/svg+xml;utf8,%3Csvg%20xmlns%3D%22http%3A//www.w3.org/2000/svg%22%20width%3D%228%22%20height%3D%2214%22%3E%3Cpath%20fill%3D%22none%22%20stroke%3D%22white%22%20stroke-width%3D%222px%22%20d%3D%22M2%202L6%207L2%2012%22/%3E%3C/svg%3E");
```
vabarbosa, line 412: we should remove the `background-image`. it adds the `>` character which is not needed:

you can get a glimpse of it when you hover over a link with surrounding text:

image

vabarbosa (changes requested):
@frankharkins thank you! this is great. i just have one minor comment (re: background-image)

eddybrando (commented):
Thank you @frankharkins! LGTM!

The only thing we still need to check is what @vabarbosa pointed out in https://github.com/qiskit-community/platypus/pull/151#discussion_r716784292.

frontend/scss/vendor/mathigon.scss
frankharkins: ah I did wonder what that did

vabarbosa (approved):
thank you!

PR #152, opened 2021-09-27
make scripts cross-platform

vabarbosa:
as currently defined some of the NPM scripts will not run on a Windows platform.

this PR updates the scripts to be cross-platform so they can run as expected in a Windows and Unix/Mac evironment

- `rimraf` is used to delete directories
- `cross-env` is used to define environment variables


y4izus (approved):
LGTM! Thank you very much!

PR #158, opened 2021-09-29
cross platform builds

vabarbosa:
working alongside @SaraM92 we found some build related issues when working in a Windows environment. this PR updates the build scripts so they can properly run in Windows and Unix-based development environments


techtolentino (approved):
Thanks @SaraM92 and @vabarbosa 

Any thoughts on testing this, for those w/o a Windows machine 😓

vabarbosa:
@Tansito i believe you mentioned you had a virtual environment with Windows? if so, do you mind doing a quick test of these changes so we have a second pass through? no worries, if you dont. thanks!

Tansito:
I don't have it but let me try tu setup one VM. It would be nice to have one for these cases anyway.

Tansito (approved):
Tested on a [Windows VM](https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/)

PR #163, opened 2021-10-01
Fix converter not parsing % in LaTex properly

### LINKED ISSUE #135: Beta version of site isn't rendering LaTeX correctly
**Describe the issue**
I was looking at the beta version of this chapter here:
https://learn.qiskit.org/course/quantum-hardware/introduction-to-quantum-error-correction-via-the-repetition-code

and it renders as such (missing the % sign):
image

**Describe what you expect to see**

It should render as shown correctly in the classic version of the chapter here:
https://qiskit.org/textbook/ch-quantum-hardware/error-correction-repetition-code.html

image

muHashh:
Fixes #135 

**before**

image

**after**

image

### DIFF
--- a/converter/textbook-converter/textbook_converter/TextbookExporter.py
+++ b/converter/textbook-converter/textbook_converter/TextbookExporter.py
@@ -216,11 +216,6 @@ def handle_markdown_cell(cell, resources, cell_number):
     headings = []
 
     for count, line in enumerate(lines):
-        if "\\%" in line:
-            l = line.replace("\\%", "\\\\%")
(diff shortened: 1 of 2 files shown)


vabarbosa (approved):
@muHashh thank you very much! this is greatly appreciated.
one thing i did is move your code further down in the function because if we have a line with inline code block  and latex percent sign ... for example,

```
this line has `code block` and $100\%$
```

the code block would not be processed properly since your update continues the loop and does not get to the part that handles inline code.
otherwise, nice work. thank you

PR #165, opened 2021-10-05
repoint binder

### LINKED ISSUE #148: [Introduction/visualizing-entanglement] hello_qiskit not found
Running the first cell on the [visualizing entanglement page](https://learn.qiskit.org/course/introduction/visualizing-entanglement) raises this error:

```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
/tmp/ipykernel_47/3565568075.py in <module>
----> 1 from hello_qiskit import run_puzzle

ModuleNotFoundError: No module named 'hello_qiskit'
```

frankharkins:
Fixes #148

It seems thebe is ignoring the 'path' and just starting the kernel at the root. This PR works around this by moving the required files to the root too.

### DIFF
--- a/server/includes/thebe-config.html
+++ b/server/includes/thebe-config.html
@@ -1,8 +1,8 @@
 <script type="text/x-thebe-config">{
   requestKernel: true,
   binderOptions: {
-    repo: "qiskit-community/qiskit-textbook",
-    ref: "platypus-binder",
+    repo: "qiskit-community/platypus",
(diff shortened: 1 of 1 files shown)


frankharkins:
Maybe this is something we should try: https://discourse.jupyter.org/t/tip-speed-up-binder-launches-by-pulling-github-content-in-a-binder-link-with-nbgitpuller/922

PR #166, opened 2021-10-05
User progress page

korgan00:


frontend/ts/courses.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (!courseList) {
+    courseList = (loadYAML(document.getElementById('toc')?.textContent || '') || []) as [Course]
+    courseList.forEach((course) => {
+      course.id = course.url.startsWith('/') ? course.url.substring(1) : course.url
+      course.sections.forEach((section) => {
+        section.url = `/course/${course.id}/${section.id}`
+        section.progress = getUserData()?.[course.id]?.[section.id]?.progress
+      })
+    })
+  }
```
vabarbosa, line 30: from my understanding, `getUserData()` this call could be moved outside of the nested `forEach` statements so it is not called repeatedly. i.e.,

```suggestion
  if (!courseList) {
    const userData = getUserData()
    courseList = (loadYAML(document.getElementById('toc')?.textContent || '') || []) as [Course]
    courseList.forEach((course) => {
      course.id = course.url.startsWith('/') ? course.url.substring(1) : course.url
      course.sections.forEach((section) => {
        section.url = `/course/${course.id}/${section.id}`
        section.progress = userData?.[course.id]?.[section.id]?.progress
      })
    })
  }
```

frontend/vue/components/common/ColumnFlowGrid.vue
```diff
@@ -0,0 +1,47 @@
+<template>
+  <div
```
techtolentino, line 2: thank you for creating this component 💪🏽

korgan00:
> @korgan00 thank you!
> 
> i had one comment (below)... but then started to think about this some more and wanted to get your thoughts.
> 
> all the information (courses, sections, urls, etc) are available on the server side. couldnt we just prep the full data in the server and not have to duplicate/similar code in the frontend? that would also address the translations as we would get the appropriate/translated titles from the server (with fallback to English if language not available).
> 
> for example, @mathigon/studio has an endpoint that returns all the courses
> 
> * https://github.com/mathigon/studio/blob/main/server/serve.ts#L15
> 
> what are your thoughts, am i missing or misunderstanding something?

I think you are right. I would like to have a conversation about that and try to unify some architectural things like having the interfaces accessible from both sides of the code (frontend and server) and some other stuff.

I will move the logic to the server at this moment to prevent code duplication.

PR #181, opened 2021-10-21
update slack invite link

frankharkins:


frankharkins:
Yes I can do that. Can I do that in a separate PR though? The current Slack link has expired and I won't be able to take a look at the css until next week.

vabarbosa (approved):

PR #184, opened 2021-10-22
QML Approaches tab panel

korgan00:
fix https://github.com/Qiskit/qiskit.org/issues/2232


vabarbosa (approved):
@korgan00 thank you! i am going to merge this and we can follow with separate PRs after discussions around different size viewports, etc

PR #195, opened 2021-11-02
Handle first markdown heading

techtolentino:
Fixes: https://github.com/Qiskit/qiskit.org/issues/2241

**Notes**
This is also related to the Mathigon Studio `renderer.js` file. 

**Why is this issue happening?**
The renderer used in Mathigon treats _numbered headings_ as a listitem, seen [here](https://github.com/mathigon/studio/blob/main/build/markdown/renderer.js#L78) - this shows _why_ these headers are being returned early, wrapped in `<li>` elements.

This fix is to use traditional HTML formatting, to handle this.

<br/>
<br/>

**Screenshot showing listitem test class rendering in DOM**
Screen Shot 2021-11-02 at 3 09 04 PM

<br/>
<br/>

**Screenshot showing heading rendered w/ HTML syntax**
Screen Shot 2021-11-02 at 3 19 44 PM


vabarbosa (changes requested):
@techtolentino as discussed i think we can make the change in the converter:

https://github.com/qiskit-community/platypus/blob/main/converter/textbook-converter/textbook_converter/TextbookExporter.py#L192

so we dont have to require all notebooks to get updated

PR #196, opened 2021-11-03
Fix first headers bug

techtolentino:
Fixes: https://github.com/Qiskit/qiskit.org/issues/2241

**showing before and after**
Screen Shot 2021-11-03 at 10 06 35 AM


vabarbosa (approved):
@techtolentino thank you!

PR #199, opened 2021-11-05
Disable preview for drafts and forks

Tansito:
Fix https://github.com/Qiskit/qiskit.org/issues/2276

The idea with this PR is to disable the preview-ga for forks and drafts to prevent false errors in the PR. Initially this PR is for platypus but I will open the same in qiskit.org as soon as this PR is approved.


vabarbosa (approved):
thank you!

PR #204, opened 2021-11-11
Fix preview deployments

Tansito:
Fix https://github.com/Qiskit/qiskit.org/issues/2298

Instead of updating the application that in code engine generates some problems, we delete the application and we create it again now.


Tansito:
After talking with Va there was a misunderstood and code engine works fine so I proceed to close this one.

PR #205, opened 2021-11-11
Added a step to prune untagged images

Tansito:
Fix https://github.com/Qiskit/qiskit.org/issues/2298


abdonrd:
https://github.com/qiskit-community/platypus/runs/4179971188?check_suite_focus=true#step:8:17

>Listing images...
>
>FAILED
>You are not authorized to access the specified resource.

Tansito:
Yes, I talked with @vabarbosa about this, we need to check permissions before continue.

vabarbosa:
> Yes, I talked with @vabarbosa about this, we need to check permissions before continue.

it's been added to the shared vault

vabarbosa:
@Tansito thinking about this some more and yes, there is a permission issue we need to resolve. however, we should be able to address the initial issue with the current permissions we have.

using `image-prune-untagged` requires more permissions because we are trying to remove any untagged image (rather then just the image we created). this may not be the best approach since we could inadvertently remove other images. i don't expect it, but there could be other uses for the repository in the future and we don't want to delete all untagged images. 

i submitted an update to this PR which checks if an images already exists for this PR and removes only that image before creating the new one.  hopefully, this should prevent the PR updates from leaving behind untagged images. this also helps with the space limitation as we delete the previous image (giving us some space back) before creating the new one.

if i can get a second pair of eyes to review & make sure i did mess anything up (or maybe you have a better approach), it would be appreciated.

Tansito:
Sounds good, I will try to invest some time on this today to review your last commits then, thanks @vabarbosa 💯

Tansito:
The changes makes sense @vabarbosa , I will start migrating this PR to the other repositories too. Thank you!

techtolentino (approved):
reviewed this w/ Va on a call...

approving the pruning 🚢

abdonrd (approved):
Good job! 👏🏻

PR #210, opened 2021-11-18
Update side-panel default view

techtolentino:
Fixes https://github.com/Qiskit/qiskit.org/issues/2095

-  Moved notation and vocabulary data to a higher component, and passing data down to child components
- if added data has no `length`, choose the `Glossary` panel as the default view
- `EmptyState` is still visible if there are no `LessonNotes`


vabarbosa (approved):
thank you!

PR #216, opened 2021-11-30
Carbon Mathigon Dropdown Style

muHashh:
This PR overrides the mathigon style of the multi-choice dropdowns, in order to align with Carbon's design principles.

Wrong and right answers:
image

Dropdown:
Desktop Screenshot 2021 11 30 - 11 14 00 78 (2)

Fixes [#2092](https://github.com/Qiskit/qiskit.org/issues/2092).


vabarbosa (changes requested):
@muHashh this is great. thanks~

if possible can you replace the wrong answer icon. i think the [**Error filled** icon](https://www.carbondesignsystem.com/guidelines/icons/library/):

image

is more accurate to the [Figma design](https://www.figma.com/file/xh0oqTfxDeCMsMmoBAYHWW/Qiskit-components?node-id=395%3A891) than the **Critical** icon.


also did the `package-lock.json` need to be changed?

vabarbosa:
@JRussellHuffman quick question regarding the error color. should it be consistent/same color across the various components? the reason i ask is because looking at the [components Figma file](https://www.figma.com/file/xh0oqTfxDeCMsMmoBAYHWW/Qiskit-components?node-id=130%3A0), for this component it is set to RED/60

image

but for the other components it is set to RED/50

image

techtolentino (commented):
@muHashh - nice job 👏🏽 

Aside from Va's comments, the only part I'm seeing is that the `height` of the fields, in the Figma spec, shows `24px`, and in the Textbook, the elements are taller.

Design shows `24`
Screen Shot 2021-11-30 at 7 06 15 AM


Live Textbook shows `26` height
Screen Shot 2021-11-30 at 7 05 07 AM

`.choice` elements are `34px` tall... 
Screen Shot 2021-11-30 at 7 05 20 AM


According to the design, they should be `24`... is that too tight?

muHashh:
> @muHashh this is great. thanks!
> 
> if possible can you replace the wrong answer icon. i think the [**Error filled** icon](https://www.carbondesignsystem.com/guidelines/icons/library/):
> 
> image
> 
> is more accurate to the [Figma design](https://www.figma.com/file/xh0oqTfxDeCMsMmoBAYHWW/Qiskit-components?node-id=395%3A891) than the **Critical** icon.
> 
> also did the `package-lock.json` need to be changed?

Thanks Va. 

I have updated it so that it's the proper SVG, and the `package-lock.json` file does not need to be changed so the changes can be ignored.

techtolentino:
> Thanks Randy.
>
> I felt 24px was a bit too too small, expecially on Mobile, so that's why it's taller. If it needs to be 24px I could change it so that it is.


@muHashh - I agree, after seeing the `24` on the live page, it looked like it could have usability issues. I like the spacing you gave it here... 

@JRussellHuffman - CC ☝🏽

JRussellHuffman:
Regarding the red color, it should be red-50, which will make it accessibility compliant. I'll update the design file. 

Regarding the 24px issue, can you share a preview link so I can also see it on a live page?

vabarbosa:
@muHashh i merging this into forked branch (instead of clone) so we can get previews.

vabarbosa:
continued in https://github.com/qiskit-community/platypus/pull/218

PR #223, opened 2021-12-02
Fix misspelling of "performs" each time it appears

garrison:


eddybrando (approved):
Thank you very much, @garrison!

PR #244, opened 2022-01-11
Updated preview workflow with the new reusable workflow logic

Tansito:
This PR will replace the old github action preview logic for the new reusable workflow that we are applying in [qiskit.org](https://github.com/Qiskit/qiskit.org). As reference we have the issue https://github.com/Qiskit/qiskit.org/issues/2398

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,118 +1,16 @@
-name: PR Preview
+name: Preview
 
 # run when a pull request is made
 on: [pull_request]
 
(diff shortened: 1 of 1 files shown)


Tansito:
How in this case `platypus` seems to use the PORT 8080 we can just remove that variable to use the default value.

abdonrd (approved):
Yay! 🎉

vabarbosa (approved):
LGTM!
thank you.

Issue #265, opened 2022-01-14
Update video on login page


techtolentino:
- [x] Update video on the login page. We created a new video that is a little cleaner and includes QML showpieces 
- [x] make video take up entire bottom section

Note: @JRussellHuffman will have the new video

Design file [here](https://www.figma.com/file/SOnOgCGQIWx4oAlSFhIsIX/1972-User-account?node-id=1114%3A7615)

vabarbosa:
resolved by #272

PR #272, opened 2022-01-14
Update `/signin` page video

techtolentino:
Fixes https://github.com/Qiskit/platypus/issues/265

- [x] added new video asset
- [x] updated dom structure and responsive styling
- [x] cleaned up unused code

Branch preview - https://platypus-pr-272.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/signin


vabarbosa:
@techtolentino thank you!

one comment, depending on the browser size the video gets cut-off. since the page doesn't scroll, it can look a bit off:

**full video**

https://user-images.githubusercontent.com/13156555/149579972-ab65dbdc-9d31-4095-b927-39eca523b5f6.mov


**cut-off video**


https://user-images.githubusercontent.com/13156555/149579996-d8a722d9-5932-47ab-9335-8a3a9085ce62.mov

techtolentino:
@JRussellHuffman - 👋🏽, can you please review this from a design standpoint?, mainly, wanting to get your feedback on the responsiveness.

As Va pointed out, at certain screen sizes, the video cuts off. What we did here is allow scrolling within that left section, if the viewport is too small to allow the video to be seen, fully.

thanks 🙏🏽

JRussellHuffman:
The left hand section scrolling is a good solution. I'm good with it 👍 
I'm testing it at other responsive sizes, and sometimes the left scrolling works and other times it doesn't. I'm not sure if that's a nuance of the preview link, but for example, I can't scroll with a screen size of 1024 x 532, but I can at 750 x 532.

techtolentino:
@JRussellHuffman - you're right about the limitations of the scrolling. You can see [here](https://github.com/Qiskit/platypus/pull/272/files#diff-fbcc788d3c001475f9c782f39c26ef8d0f4e77c7de4ab65b2721cefadf4400adR92), that I placed the scroll condition to only happen _up until_ the large breakpoint, `1056 px` in width.

In your opinion, should we allow the scroll at all times?

JRussellHuffman:
Is there a benefit to preventing scroll at larger sizes? It seems like we should allow it at all sizes. Or to say it differently, we should allow it in any scenario where there is content beyond the size of the screen (which seems to happen at specific points in large and small breakpoints).

techtolentino:
> _Is there a benefit to preventing scroll at larger sizes?_ 

@JRussellHuffman - you're right, there's no _benefit_ to preventing scrolling, other than it fitting nice and snug, all on one screen. However, given that we should account for all screens, I've removed the breakpoint media-query, and if that section needs to scroll, it will now.

JRussellHuffman:
Great. I'm testing it now and it still doesn't scroll for me after hard reload. I'm guessing the update isn't pushed yet. Let me know and i'll take one last look.

JRussellHuffman:
Page is updated with the new scrolling. Everything looks good to me 👍

vabarbosa (approved):
thank you

PR #273, opened 2022-01-17
Prevents GitHub Actions to be executed by forks

abdonrd:
Also parity naming with Qiskit/qiskit.org: https://github.com/Qiskit/qiskit.org/pull/2433

### DIFF
--- a/.github/workflows/preview-cleanup.yml
+++ b/.github/workflows/preview-cleanup.yml
@@ -1,6 +1,5 @@
 name: Preview
 
-# run when a pull request is closed
 on:
   pull_request:
     types: [closed]
--- a/.github/workflows/preview-cleanup.yml
+++ b/.github/workflows/preview-cleanup.yml
@@ -6,9 +6,9 @@ on:
 
 jobs:
   code-engine:
-    # Run the job if it's not triggered by a PR or if the PR is from the same repository
-    # This is because this job requires a secret, and secrets cannot be shared with forks
-    if: github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository
+    # Run only from the original repository
+    # Because this job requires secrets, which cannot be shared with forks
+    if: github.event.pull_request.head.repo.full_name == github.repository
 
     uses: Qiskit/gh-actions/.github/workflows/code-engine-cleanup.yml@main
     with:
--- a/.github/workflows/preview-cleanup.yml
+++ b/.github/workflows/preview-cleanup.yml
@@ -7,6 +7,10 @@ on:
 
 jobs:
   code-engine:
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -13,22 +11,23 @@ on:
 
 jobs:
   build-test-release:
+    # Run the job if the PR is from the same repository
```
Tansito, line 14: Just one comment here, how this github action is not fired from a PR should we do this check? And if so, should we do the same in qiskit.org? https://github.com/Qiskit/qiskit.org/blob/main/.github/workflows/deploy.yml#L14

Tansito (changes requested):

.github/workflows/release.yml
abdonrd: Because the `on` trigger is set for `workflow_dispatch`, `push` and `schedule`.
Not for `pull_request`.

Tansito (dismissed):

Tansito:
LGTM!

.github/workflows/preview-cleanup.yml
```diff
@@ -1,12 +1,15 @@
...
-# run when a pull request is closed
 on:
   pull_request:
     types: [closed]
 
 jobs:
   code-engine:
+    # Run the job if it's not triggered by a PR or if the PR is from the same repository
+    # This is because this job requires a secret, and secrets cannot be shared with forks
+    if: github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository
```
vabarbosa, line 11: just curious, since this is only `on: pull_request` is there a case when `github.event.pull_request == null`?

.github/workflows/preview-deploy.yml
```diff
@@ -1,12 +1,14 @@
...
+on:
+  pull_request:
 
 jobs:
   code-engine:
-    # Prevents action to be executed by forks
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    # Run the job if it's not triggered by a PR or if the PR is from the same repository
+    # This is because this job requires a secret, and secrets cannot be shared with forks
+    if: github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository
```
vabarbosa, line 10: same question, since this is only `on: pull_request` is there a case when `github.event.pull_request == null`?

vabarbosa (approved):
looks good. thank you.
just had one question

.github/workflows/preview-cleanup.yml
abdonrd: You're right! No needed in this case!

eddybrando (approved):
LGTM! Thanks, @abdonrd!

PR #276, opened 2022-01-17
Fix release.yml quotes

abdonrd:
Seems like double quotes doesn't work on the `if` condition:

https://github.com/Qiskit/qiskit.org/actions/runs/1708639108


vabarbosa (approved):
good to know. thanks

PR #296, opened 2022-01-19
dependencies update

vabarbosa:
- update dependencies
- install production only dependencies for preview


techtolentino (approved):
Va, I checked out the branch and everything seems to be working correctly.

Not sure how much more I can review, but the icons, colors, components all seem to be working as expected. 

LGTM 🚀

Issue #280, opened 2022-01-18
Investigate why platypus is breaking from a clean installation


Tansito:
From a clean project doing a `npm install` generates errors in the front. We need to identify which dependency is creating the problem.

techtolentino:
@Tansito - I'm curious, what errors do you see on the front end? 

I _am_ seeing this error (missing CTA), but I'm not sure if this is what you're referring to.
<img width="1644" alt="Screen Shot 2022-01-18 at 1 09 34 PM" src="https://user-images.githubusercontent.com/6276074/150019420-8d24bc34-0de1-4dc1-b824-f40d1cb2f3cd.png">

Tansito:
Yeah, the error is `isExternal is not a function` that I think is generating the error in the test `1) next and previous links navigate to next and previous pages`. Only happens when you run `npm install` but not `npm ci`. We installed the renovate app to see if it was one of our dependencies but not, so a I got a bit stuck at this point.

Tansito:
Talking with @korgan00 it seems that we can try to do a workaround moving the static method from the class to an export method.

vabarbosa:
@Tansito @korgan00 i did some more investigation and the issue is with this `vue-loader-v16` which is  dependency of `@vue/cli-service`.

more specifically `@vue/cli-service` has a dependency of:

```json
"vue-loader-v16": "npm:vue-loader@^16.1.0"
```

currently, the `package-lock.json` includes the info for version `16.8.1`:

```json
        "vue-loader-v16": {
          "version": "npm:vue-loader@16.8.1",
          "resolved": "https://registry.npmjs.org/vue-loader/-/vue-loader-16.8.1.tgz",
          "integrity": "sha512-V53TJbHmzjBhCG5OYI2JWy/aYDspz4oVHKxS43Iy212GjGIG1T3EsB3+GWXFm/1z5VwjdjLmdZUFYM70y77vtQ==",
```

so when you run `npm ci` it installs that version and everything works as expected.

however, depending on the environment when you run `npm install` it installs `16.8.3`:

```json
        "vue-loader-v16": {
          "version": "npm:vue-loader@16.8.3",
          "resolved": "https://registry.npmjs.org/vue-loader/-/vue-loader-16.8.3.tgz",
[...]

PR #301, opened 2022-01-20
move static methods out of component class

vabarbosa:
workaround for https://github.com/Qiskit/platypus/issues/280

this PR moves the static methods out of the `BasicLink` component class


frontend/vue/components/constants/helper.ts
```diff
@@ -0,0 +1,16 @@
+
```
Tansito, line 1: Thanks for this @vabarbosa , what do you think moving this file to somewhere like: `frontend/vue/components/helpers/link.ts` because from `constants/helper.ts` could be a bit confusing due to they are methods not constants themselves. 

Apart from that I tested that it's working with new dependency versions so it seems that this could work 🎉

Tansito (approved):

frontend/vue/components/constants/helper.ts
korgan00: We have a file called `utilities.ts` into the `frontend/ts` I think it would be a good place to move this code.

korgan00 (approved):

PR #308, opened 2022-01-20
Update renovate

vabarbosa:
related: #285 
this PR sets a schedule and limitations for renovate:

- run only weekends 
- and create only 5 PRs at a time

these settings are up for discussion so feel free to recommend/suggest a better configuration

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,5 +5,9 @@
     ":separateMultipleMajorReleases",
     ":maintainLockFilesWeekly"
   ],
-  "rangeStrategy": "bump"
+  "rangeStrategy": "bump",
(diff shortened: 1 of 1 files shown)


vabarbosa:
merging this so we can slow down all the renovate PRs but we should have discuss on what is a good set of configuration

abdonrd:
It's fine with me now that there are so many outdated dependencies. But when everything is more or less up to date I think we could remove these restrictions.

vabarbosa:
> It's fine with me now that there are so many outdated dependencies. But when everything is more or less up to date I think we could remove these restrictions.

yeah, we can definitely reset it once we've worked through all the current dependencies and get to a manageable status

Issue #342, opened 2021-12-02
Segment "Loaded a Page" event not logged on all course pages


pandasa123:
### Steps to reproduce the problem

I believe "Loaded a Page" is a standard event if the analytics script is initialised. Here are some pages where this event isn't logged

- [Introduction](https://learn.qiskit.org/course/introduction/why-quantum-computing)
- [Algorithms](https://learn.qiskit.org/course/ch-algorithms/quantum-circuits#why-classical)

### What is the current behavior?

[If we visit Amplitude, all pages under `/course` path are missing](https://analytics.amplitude.com/ibm/chart/new/0l6gfoz) from data

### What is the expected behavior?

We should be able to track every page load similar to other qiskit.org pages

PR #343, opened 2022-01-24
track page on load

vabarbosa:
Fixes https://github.com/Qiskit/platypus/issues/342

this PR updates the analytics to properly capture page access/load event

### DIFF
--- a/frontend/textbook.ts
+++ b/frontend/textbook.ts
@@ -65,9 +65,12 @@ textbook.runAfterDOMLoaded(() => {
   initIndexHighlight()
 
   // set up & trigger segment
-  initAnalytics(window.textbookAnalytics.key, window.textbookAnalytics.url)
-  textbook.trackClickEvent = trackClickEvent
-  textbook.trackPerformedSearch = trackPerformedSearch
-  textbook.trackUpdatedObject = trackUpdatedObject
-  trackPage(window.location.pathname)
+  initAnalytics(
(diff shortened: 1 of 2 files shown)


frontend/textbook.ts
```diff
@@ -65,9 +65,12 @@ textbook.runAfterDOMLoaded(() => {
   initIndexHighlight()
 
   // set up & trigger segment
-  initAnalytics(window.textbookAnalytics.key, window.textbookAnalytics.url)
-  textbook.trackClickEvent = trackClickEvent
-  textbook.trackPerformedSearch = trackPerformedSearch
-  textbook.trackUpdatedObject = trackUpdatedObject
-  trackPage(window.location.pathname)
+  initAnalytics(
```
techtolentino, line 68: @vabarbosa - this is awesome, thank you! 

I was wondering if you've been able to confirm that this'll be tracking correctly, after this change?

vabarbosa: i am not sure if we can get tracking from localhost or from the preview link since the domains are not what is expected. we may need to get confirmation from Anne once deployed. i will do a little digging around and see if there is way to confirm.

techtolentino: Okay, thanks for the response, I don't think we need to block this. Let's confirm once it's deployed to ensure we're capturing what we need.

🚢

techtolentino (approved):

PR #348, opened 2022-01-26
Add amplitude addition widget to intro course

frankharkins:
Final piece of the puzzle- added the amplitude addition widget to the intro course.


vabarbosa (approved):
thank you @frankharkins

PR #349, opened 2022-01-27
Remove '>' from filenames for windows users

frankharkins:
Windows users cannot clone the repo as `>` is invalid in windows files.


vabarbosa (approved):
thank you @frankharkins

PR #692, opened 2022-02-03
Update EULA login flow w/ route logic

### LINKED ISSUE #678: User account: update login flow w/ EULA page [logic/routing]
This is the 2nd part of #267 

The task here is to update the logic/routing for the EULA login flow
- [ ] ensure that the `[acceptedPolicies](https://github.com/mathigon/studio/blob/main/server/models/user.ts#L41)` property is added to the user's data object
- [ ] once the user has accepted the EULA by clicking on "Accept & Continue", they'll be directed to the `/account` page
- [ ] if a user has already `acceptedPolicies`, they should skip the EULA page and go straight to their `/account` view

techtolentino:
## Changes

Implemented the functionality for properly routing a user during the sign-in / eula process
Closes #678 

## Implementation details
- updated acceptance string
- added checks based off of the `user` data object

## How to read this PR
- if a user attempts to `/signin` but has not `acceptedPolicies`, they'll be redirected to the `/eula` page
- if a user clicks on "Continue & Accept", there's a `POST` request to accept policies


server/app.ts
```diff
@@ -60,7 +60,7 @@ const getCourseData = async function (req: Request) {
 new MathigonStudioApp()
   .get('/health', (req, res) => res.status(200).send('ok')) // Server Health Checks
   .secure()
-  .setup({ sessionSecret: 'project-platypus-beta' })
+  .setup({ sessionSecret: 'project-platypus-beta', csrfBlocklist: ['/profile/accept-policies'] })
```
Tansito, line 63: Apart from this that I'm not 100% confident with this approach I'm fine with the rest of the things. I can open an issue to investigate it a bit more but by now I think we can go with it.

Tansito (approved):

server/app.ts
Tansito: Thanks @techtolentino I will edit it with more information.

PR #701, opened 2022-02-07
add venv files to .gitignore

declanmillar:
## Changes

Adds common `venv` environment files and directories to the `.gitignore` for those wanting to use a virtual python environment for local development without the risk of accidentally adding these to the repo or cluttering up their changes.


Tansito (approved):
Thanks for this @declanmillar

PR #708, opened 2022-02-08
docs: development builds redirection

### LINKED ISSUE #685: Add information about redirection from index page to LP
The index page of Platypus redirects to https://qiskit.org/textbook-beta/.

This is intended for production as we want https://learn.qiskit.org/ to redirect to that page, where we have a better landing page that the one we could have created within the framework of Platypus.

This is causing an issue for developer experience, since this redirection is not expected unless one knows about it.

When developing locally, http://localhost:8080/ would redirect automatically to https://qiskit.org/textbook-beta/.

The same happens for all PR preview links.

We need to add some documentation information so that the developer can know that this redirection is expected and also some information about how to use the PR preview links or the local build (which is by adding a more specific path and not visiting the index page).

eddybrando:
## Changes

Added information about the development and PR preview builds' redirection.

Closes #685

## Implementation details

I tried to make the changes to the documentation very minimal, avoiding creating new sections. The hypothesis being that the longer the README file, the less likely it is for contributors to really pay attention to its information or even to read it at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,14 +42,17 @@ $ npm start
 
 > _To watch changes and rebuild automatically run with `npm run watch` after installing the dependencies._
 
-After the application has started, access the site at: http://localhost:8080/
+After the application has started, the site will be available at http://localhost:8080/, but the index route will redirect to an external landing page.
+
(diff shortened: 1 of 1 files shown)


abdonrd (approved):

Tansito (approved):
LGTM!

Issue #705, opened 2022-02-08
Sass division deprecation warnings


techtolentino:
### Steps to reproduce the problem

1. Build the app in your local terminal
2. See warnings

### What is the current behavior?

When building the application locally, we see the deprecated warnings in the console

### What is the expected behavior?

We shouldn't see these warnings anymore

### Screenshots
![Screen Shot 2022-02-08 at 7 00 49 AM (2)](https://user-images.githubusercontent.com/6276074/153015344-c81b69a1-78a5-435e-8e57-ae8e3ae32de4.png)

PR #742, opened 2022-02-17
Updated every division with the new sass syntax

korgan00:
## Changes

Fixes https://github.com/Qiskit/platypus/issues/705


eddybrando (approved):
Nice! Thank you @korgan00! 💪

Issue #271, opened 2022-01-14
Privacy page - delete data


techtolentino:

techtolentino:
@JRussellHuffman - similar here, please see the design needs from above; we need to know what the user might see/experience when deleting their account...

Thanks

JRussellHuffman:
Do we have an idea of what certain parts of the data need to be deletable? If not, we could use a generic pattern like this:
<img width="516" alt="Screen Shot 2022-01-25 at 5 09 29 PM" src="https://user-images.githubusercontent.com/7633881/151075076-b4c9ee30-d8b4-413f-a875-63ec934b4fe7.png">
 
"select all" would effectively be the same as deleting the entire account.

Tansito:
The data that we would store for a standard user:

```
{
    "_id" : ObjectId("xxxxxxxxxxxxxxxxxxxxxxxx"),
    "email" : "user@email.com",
    "previousEmails" : [],
    "firstName" : "David",
    "lastName" : "García Valiñas",
    "type" : "student",
    "country" : "US",
    "picture" : "",
    "oAuthTokens" : [ 
        "ibm:XXXXXXXXXX"
    ],
    "createdAt" : ISODate("2021-12-14T14:19:27.818Z"),
    "updatedAt" : ISODate("2022-01-17T15:50:15.987Z"),
    "__v" : 0,
    "lastOnline" : ISODate("2022-01-17T00:00:00.000Z")
}
```

Apart from that a user has a relationship with `loginanalytics`, `progress` and `session`. Tables that store when the user makes a login acction, the progression of the user in the different courses and the session of the user respectively. 

If a user wants to delete  its account the current logic marks that user with a new param called `deletionRequested` to `true` and 7 days **after** he request his deletion **ONLY** the data from `user` is deleted but the data from `loginanalytics`, `progress` and `session` continues being there.

@lerongil @pandasa123 @JRussellHuffman I hope this information can help us to advance in the GDPR to understand if we are going to need to modify something or this is enough.

JRussellHuffman:
Let me make sure I understand. Everything in the json above is the `user` material, but not `loginanalytics`, `progress` and `session`, right?
The discussion is about deleting certain parts of the data. Would they be able to delete some material without `user`, or just the entire thing? And would we need to surface the `loginanalytics`, `progress` and `session`data to be deleted as well?

[...]

PR #746, opened 2022-02-21
Junk data for a deleted user

### LINKED ISSUE #743: User account: Privacy page - delete account [logic]
This issue is for updating the modal `Delete account` button (danger themed) with the full deletion logic. We needed to separate this from the UI pull request since _that_ PR was ready for review, and there was still some research that needed to be done, for this logic story.

Related to #271

Tansito:
Fix #743 

The idea for this PR is complete the logic for a deletion account process defined in this issue: https://github.com/Qiskit/platypus/issues/271

The way that we are doing it is generating a random identifier with [nanoid](https://www.npmjs.com/package/nanoid). We modify the original private data with junk data.


frontend/vue/components/UserAccount/DeleteUserDataSection.vue
```diff
@@ -93,8 +93,9 @@ export default defineComponent({
       this.isButtonDisabled = true
       this.isModalVisible = true
     },
-    modalDeleteAction() {
-      this.isModalVisible = false
+    modalDeleteAction(ev: any) {
+      ev.preventDefault()
+      window.location.href = '/delete/account'
```
Tansito, line 98: @techtolentino @korgan00 I think I will need some help here. My idea was to modify this button with a link to make use of the redirection from the end-points but I couldn't find an element that fits correctly in the modal so I just opted for this option but probably you have a better approach.

server/app.ts
```diff
@@ -88,6 +89,23 @@ new MathigonStudioApp()
...
     res.json(courses)
   })
+  .get('/delete/account', async (req, res) => {
+    if (!req.user) return {error: 'unauthenticated', errorCode: 401, redirect: '/signin'}
+
+    const alphabet = 'abcdefghijklmnopqrstuvwxyz'
+    // 21 value is to maintain the probability collision similar to UUIDv4
+    const nanoid = customAlphabet(alphabet, 21)
+    const randomString = nanoid()
+    req.user.email = `deleted-${randomString}@qiskit.com`
```
techtolentino, line 99: I know this is a fake email -- so I'm wondering if the `@qiskit.com` part matters... or should we be using `@qiskit.org`

techtolentino (approved):
LGTM, thanks @Tansito 🚀

PR #757, opened 2022-02-22
Remove unused ESLint dependencies

abdonrd:
## Changes

Remove unused ESLint dependencies.

Right now we just use the `@nuxtjs/eslint-config-typescript`:

https://github.com/Qiskit/platypus/blob/fda6e21632e92868c9ff0f561f65de205e354524/.eslintrc.js#L2-L4

And it has all these removed dependencies inside.


Tansito:
🤞

PR #758, opened 2022-02-22
Lock generation for npm 8

Tansito:
Related with https://github.com/Qiskit/qiskit.org/issues/2504

We are updating the project to support officially node-16/npm-8.

As a note with the update we had a problem in the build step and we needed to add `vue-demi` as a dev-dependency in the project (I would like to ping here to @eddybrando)


abdonrd (approved):
🤞🏻

Issue #763, opened 2022-02-23
Syllabus: Create API endpoint to create a syllabus


eddybrando:
We need to create an API endpoint to **create** a new syllabus (i.e. store it in the database).

This includes storing in the the database the data entered by the user as well as other automatically generated data, like the **random access code**.

Issue #747, opened 2022-02-21
Syllabus: Define database structure


eddybrando:
We need to discuss and define the database architecture to support the new syllabus feature.

eddybrando:
FYI @Tansito, @korgan00

PR #786, opened 2022-03-02
Save a syllabus

### LINKED ISSUE #762: Syllabus: Create "Syllabus" object in database
We need to create the "Syllabus" object/model/schema in the database as defined in https://github.com/Qiskit/platypus/issues/747.

Tansito:
Fix #762 #763 

This PR creates a syllabus whenever you send a **POST** to `/syllabus` end-point. The structure of that body is defined here #747 .

A little example:

Screenshot 2022-03-09 at 16 23 33

The output for this end-point will be the syllabus itself plus its `id` and a `code` generated dynamically following the same structure: **XXX-XXX**.


eddybrando (approved):
Awesome job @Tansito!

This creates a very robust base to interact with our database in various ways in the future. 💪

PR #792, opened 2022-03-06
Fix typo

ArdaGurcan:
## Changes

Fixed the control qubit to be the leftmost qubit as is the case in the image.

## Screenshots
image


frankharkins (approved):
Thank you!

techtolentino (approved):
A "thank you" from @frankharkins to me means this is good to go 🚀

Issue #770, opened 2022-02-23
Syllabus: Create "Syllabus" page UI


eddybrando:
We need to create the "Syllabus" page as designed in Figma: https://www.figma.com/file/f1QMqiXSgt2FCt0itvsA24/676-Professor-Syllabus?node-id=50%3A1018

![Screenshot 2022-02-23 at 18 40 32](https://user-images.githubusercontent.com/22047320/155375888-2ab154be-d294-4adc-bfbd-83995202a103.png)

To retrieve the data, we can use the API endpoint created in https://github.com/Qiskit/platypus/issues/765.

PR #806, opened 2022-03-10
Syllabus page UI

korgan00:
## Changes

fixes https://github.com/Qiskit/platypus/issues/770

## How to read this PR

URL: `/syllabus/123-abc`

### DIFF
--- a/frontend/vue/components/Syllabus/SyllabusGeneralInformation.vue
+++ b/frontend/vue/components/Syllabus/SyllabusGeneralInformation.vue
@@ -1,22 +1,11 @@
 <template>
   <div class="syllabus-general-information">
-    <div class="syllabus-general-information__item">
-      <span class="syllabus-general-information__item__heading">Instructor:</span> <slot name="instructor" />
-    </div>
-    <div class="syllabus-general-information__item">
-      <span class="syllabus-general-information__item__heading">Location:</span> <slot name="location" />
-    </div>
-    <div class="syllabus-general-information__item">
-      <span class="syllabus-general-information__item__heading">University:</span> <slot name="university" />
-    </div>
-    <div class="syllabus-general-information__item">
-      <span class="syllabus-general-information__item__heading">Office Hours:</span> <slot name="office-hours" />
-    </div>
-    <div class="syllabus-general-information__item">
-      <span class="syllabus-general-information__item__heading">Class Hours:</span> <slot name="class-hours" />
-    </div>
-    <div class="syllabus-general-information__item">
-      <span class="syllabus-general-information__item__heading">Email:</span> <slot name="email" />
+    <div
+      class="syllabus-general-information__item"
+      v-for="(field, index) in fields"
+      :key="index"
+    >
+      <span class="syllabus-general-information__item__heading">{{ field.name }}:</span> {{ field.value }}
     </div>
   </div>
 </template>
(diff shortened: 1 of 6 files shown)


frontend/vue/components/Syllabus/SyllabusView.vue
```diff
@@ -0,0 +1,181 @@
...
+        Chapters
+      </div>
+      <ColumnFlowGrid
+        tag="ul"
+        class="syllabus-view__work-package__chapter-list"
+        :elements="workPackage.chapterList"
+      >
+        <template #default="slotProps">
+          <li class="syllabus-view__work-package__chapter-list__item">
+            <BasicLink class="syllabus-view__work-package__chapter-list__chapter" :url="slotProps.element.url">
```
techtolentino, line 51: One small thing I noticed is in the URL structure for the course links... 

I think `../syllabus/course/...` needs to be updated to remove `syllabus`

Screen Shot 2022-03-10 at 11 34 09 AM (2)

frontend/vue/components/Syllabus/SyllabusView.vue
```diff
@@ -0,0 +1,181 @@
...
+      </template>
+      <template #class-hours>
+        {{ syllabus.classHours }}
+      </template>
+      <template #email>
+        {{ syllabus.email }}
+      </template>
+    </SyllabusGeneralInformation>
+    <div
+      v-for="(workPackage, idx) in syllabus.workPackages"
```
techtolentino, line 30: interested in what is a `workPackage` 

(I'm okay w/ it, but would just like to understand what it means since I'm new to the term...)

is this like a course? module?

frontend/vue/components/Syllabus/SyllabusGeneralInformation.vue
```diff
@@ -0,0 +1,54 @@
...
+      <span class="syllabus-general-information__item__heading">University:</span> <slot name="university" />
+    </div>
+    <div class="syllabus-general-information__item">
+      <span class="syllabus-general-information__item__heading">Office Hours:</span> <slot name="office-hours" />
+    </div>
+    <div class="syllabus-general-information__item">
+      <span class="syllabus-general-information__item__heading">Class Hours:</span> <slot name="class-hours" />
+    </div>
+    <div class="syllabus-general-information__item">
+      <span class="syllabus-general-information__item__heading">Email:</span> <slot name="email" />
```
techtolentino, line 19: Question for @gracelindsell: 

Should we code the email in the syllabus view, to be a clickable link? or just plain-text is fine?

techtolentino (changes requested):
@korgan00 - this is great 👏🏽 

I just had a couple of small comments, and a question for Grace. 

awesome work Goyo

frontend/vue/components/Syllabus/SyllabusView.vue
korgan00: Good question!
I thought of a better name than a course to differentiate between our courses and their chapter selection.
The first approach was `weeks`, but it could be two weeks or a semester or maybe don't have a time constraint.
I decided to call it `workBlocks` or `workPackages` because it is an amount of work that has some coherence and should be worked "in block".
If you have any other suggestions, I will accept them.

frontend/vue/components/Syllabus/SyllabusView.vue
korgan00: Yeah, thanks!
It is an issue related with the mock data. It should be fixed now

frontend/vue/components/Syllabus/SyllabusGeneralInformation.vue
techtolentino: (updating comment to a non-blocking comment... we can always address this later)

eddybrando:
Is it just happening to me or can you also not select any text in https://platypus-pr-806.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/syllabus/123-abc?

frontend/vue/components/Syllabus/SyllabusGeneralInformation.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <div class="syllabus-general-information">
+    <div class="syllabus-general-information__item">
+      <span class="syllabus-general-information__item__heading">Instructor:</span> <slot name="instructor" />
```
eddybrando, line 4: I believe we should be using props instead of slots since we're always expecting text strings. We could then profit from type checking and also easily cover the scenarios where one of the attributes isn't defined with `v-if`.

We should then also simplify this repetitive sections with a `v-for` loop.

frontend/vue/components/Syllabus/SyllabusLayout.vue
```diff
@@ -9,23 +9,20 @@
         {{ $translate("This is placeholder text for a page where students are viewing a syllabus.") }}
       </template>
     </UserAccountSectionHeader>
-    <section class="syllabus__content">
```
eddybrando, line 12: We should also remove the CSS definition of the `syllabus_content` class since it's now no longer used.

frontend/vue/components/Syllabus/SyllabusView.vue
eddybrando: I'm also leaning towards "module" here. It's also closer to my academic experience.

@Tansito, since ideally we should use the same vocabulary here and in the database, do you think that this name makes sense for this blocks defining the workload of chapters during a specified period?

image

frontend/vue/components/Syllabus/SyllabusView.vue
```diff
@@ -0,0 +1,181 @@
...
+export default defineComponent({
+  name: 'SyllabusView',
+  components: {
+    SyllabusGeneralInformation,
+    BasicLink,
+    ColumnFlowGrid
+  },
+  data () {
+    return {
+      // TODO mock data
```
eddybrando, line 76:
```suggestion
      // TODO replace mock data
```

eddybrando (approved):
Thanks a lot @korgan00! The page is looking really great! 🙂 

I just had a couple of comments, mostly to simplify the code.

eddybrando (changes requested):
I'm sorry! I meant to mark my review as "request changes".

pandasa123:
> Is it just happening to me or can you also not select any text in https://platypus-pr-806.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/syllabus/123-abc?

I'm facing the same issue @eddybrando. Looks like `user-select: none` or something?

JRussellHuffman:
I also can't select anything

korgan00:
The text selection is another mathigon inherited thing. It should be fixed now.

PR #824, opened 2022-03-15
Find syllabi by user

### LINKED ISSUE #766: Syllabus: Create API endpoint to get the data of all syllabi owned by a logged in user
We need to create an API endpoint that retrieves the data of **all of the syllabi that are owned by a logged in user**.

We need to consider the following scenarios:

- [x] The user is not logged in
- [x] The logged in user owns no syllabi
- [x] The logged in user has 1 syllabus
- [x] The logged in user has many syllabi

Tansito:
Fix #766 

This end-point retrieves syllabus related with the user paginated. By now the end-point doesn't populate relations automatically. An example of response is:

Screenshot 2022-03-15 at 11 41 03

In case the user is not authenticated the end-points returns a 401 error.

This issue depends on #786


eddybrando (approved):
Looks great to me! Thank you @Tansito! 🙂

PR #825, opened 2022-03-16
Get syllabus by code

### LINKED ISSUE #765: Syllabus: Create API endpoint to get a syllabus' data
We need to create an API endpoint to **get** an existing syllabus' data.

Tansito:
Fix #765 

The end-point `syllabus/:code` will retrieve the information for the syllabus related with that code. Some points to take in consideration:

- The code can be written in lower case or upper case
- The user must be authenticated to use the end-point
- With the code any user can retrieve the syllabus information, we are not checking if the user should have access (this guarantee public access meanwhile you have the syllabu's code)


server/modules/syllabus/commands/find-syllabus-by-code/find-syllabus-by-code-controller.ts
```diff
@@ -0,0 +1,35 @@
...
+import { FindSyllabusByCodeHttpRequest } from './find-syllabus-by-code-dto'
+import { FindSyllabusByIdService } from './find-syllabus-by-code-service'
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const FindSyllabusByCodeController = async (req: Request, res: Response, next: NextFunction) => {
+  if (!req.user || !req.user.acceptedPolicies) {
+    const error = new UnauthorizedException()
+    res.status(error.code)
+    return res.json(error)
+  }
```
eddybrando, line 14: Currently, there is no need to restrict this functionality to authenticated users only.

eddybrando (changes requested):
Awesome job @Tansito! 💪 

I've only left one comment to remove the authentication check in the endpoint, since it's not required at the moment.

PR #828, opened 2022-03-18
Move toc to working folder

### LINKED ISSUE #783: User Account: Learning course list not showing up

1. Go to learn.qiskit.org/account (and if you don't have one, create one using IBMId)
2. Click on 'Learning'
3. See empty page

The Learning page doesn't show any courses

The Learning page should render a list of courses, along with progress icons

### Screenshots (optional but very helpful)
Screen Shot 2022-02-28 at 12 09 49 PM

### Environment
MacOS on latest Chrome and Firefox

Tansito:
Fix #783 

Investigating with @korgan00 we think that the `toc.yaml` was never being uploaded in production. With this change what we want to do is copy the file to the working folder and use that folder to access to it (as we do with the rest of the resources) to avoid future problems and not access to the notebooks folder for anything.

Co-authored by @korgan00 🎉

### DIFF
--- a/.gcloudignore
+++ b/.gcloudignore
@@ -23,7 +23,6 @@ converter/
 cypress/
 docs/
 notebooks/
-!notebooks/toc.yaml
 scripts/
 .browserslistrc
 .editorconfig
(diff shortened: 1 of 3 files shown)


eddybrando (approved):
Seems plausible 😬. I guess we need to give it a try while deployed.

PR #829, opened 2022-03-18
Update syllabus

### LINKED ISSUE #764: Syllabus: Create API endpoint to update a syllabus
We need to create an API endpoint to **update** an existing syllabus.

Tansito:
Fix #764 

The end-point POST `challenge/:id` will make an update over the specific syllabus with the data that you pass in the body.

- This end-point will need that the user is authenticated
- Only the owners of the syllabus can edit it
- In the body you only need to pass the fields that you want to change
- Initially for update the syllabus we use the `id` instead of the `code` to access to it


server/libs/database/mongoose-repository-base.ts
```diff
@@ -19,13 +19,21 @@ implements RepositoryPort<QueryParams, Domain> {
...
     return this.mapper.toDomainEntity(document)
   }
 
+  async update (id: string, data: Partial<Domain>): Promise<Domain | null> {
+    const document = await this.EntityModel.findByIdAndUpdate({ _id: id }, data, { new: true })
+
+    if (document) {
+      return this.mapper.toDomainEntity(document)
+    }
+    return document
```
korgan00, line 34: I prefer to manage errors with the `if` instead of managing the correct path.

```suggestion
    if (!document) {
      return null
    }
    return this.mapper.toDomainEntity(document)
```

server/modules/syllabus/database/syllabus-repository.ts
```diff
@@ -37,4 +37,13 @@ export class SyllabusRepository
...
     return document
   }
+
+  async updateSyllabusFilteredByOwner (id: string, owner: string, data: Partial<Syllabus>): Promise<Syllabus | null> {
+    const document = await this.EntityModel.findOneAndUpdate({ _id: id, owners: owner }, data, { new: true })
+
+    if (document) {
+      return this.mapper.toDomainEntity(document)
+    }
+    return document
```
korgan00, line 47: Same here. I prefer to manage errors with the `if` instead of managing the correct path.

```suggestion
    if (!document) {
      return null
    }
    return this.mapper.toDomainEntity(document)
```

korgan00 (changes requested):
Good job. Just an small suggestion.

korgan00 (approved):
LGTM

PR #841, opened 2022-03-23
Added CSRF implementation

### LINKED ISSUE #733: User account: Investigate csrf approach
Apart from this that I'm not 100% confident with this approach I'm fine with the rest of the things. I can open an issue to investigate it a bit more but by now I think we can go with it.

_Originally posted by @Tansito in https://github.com/Qiskit/platypus/pull/692#discussion_r807141671_

Tansito:
Fix #733 

To execute **POST** requests from the front we need to pass in the body the token with the key `_csrf`.


techtolentino (approved):
This works for me @Tansito 

👏🏽

PR #853, opened 2022-03-29
Mock data to initialise the database

### LINKED ISSUE #847: Populate database w/ mock Syllabus data for dev
In order to ensure that our data properly connects with the UI, we should have a way to test data coming from the backend. 

This goal is to populate the database with a few examples of syllabus data so that, as we replace the frontend mock data, we know that we've prepared for the right data structures, syllabi objects, properties, etc.

Tansito:
Fix #847 

This PR will try to populate with mock data the database for each environment that differs from `production`.

There is a new variable from `config.yaml` called `mockData.email` that you can override with your email. That way once time you log in locally you are always to have syllabi assigned to that user.


korgan00 (approved):
LGTM

PR #855, opened 2022-03-30
merge feature/professor-syllabus

korgan00:
Fixes https://github.com/Qiskit/platypus/issues/877 (keeping as draft until we get go ahead to merge)

Preview: https://platypus-pr-855.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/account#Classroom

## Changes

Completes Syllabus v1 Milestone

 
## Screenshots

image

image

image

image


vabarbosa:
@eddybrando @pandasa123 i believe this is ready to be merged. i am not sure if there was anything else being waiting on (the issues in the ["Syllabus v1" milestone](https://github.com/Qiskit/platypus/milestone/1) have been addressed). i didnt want to merge w/o first running by you to make sure i am not missing anything (after being away)

you can review result using the  [preview deployment](https://platypus-pr-855.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/account#Classroom) (sign-in with IBMid)

pandasa123:
Yep ready whenever

frontend/vue/components/Syllabus/SyllabusView.vue
```diff
@@ -0,0 +1,142 @@
...
+      :key="`${idx}-${course.title}`"
+      class="syllabus-view__course"
+    >
+      <div class="syllabus-view__course__title">
+        {{ course.title }}
+      </div>
+      <p class="syllabus-view__course__description">
+        {{ course.description }}
+      </p>
+      <div class="syllabus-view__course__units-title">
```
pandasa123, line 21: We shouldn't show **Chapters** if there are no units

Suggested implementation:
- Adding `v-if`
- If `course.unitList` isn't empty, then show **Chapters**

frontend/vue/components/Syllabus/SyllabusFormModule.vue
```diff
@@ -0,0 +1,251 @@
...
+      saveSyllabusModuleLink: {
+        label: 'Save content',
+        url: '#'
+      },
+      syllabusSaved: false
+    }
+  },
+  mounted () {
+    getCourseList().then((courses) => {
+      this.courseList = courses.filter(course => course.type === 'learning-path')
```
vabarbosa, line 112: @korgan00 it looks like we are filtering the courses and only showing those that are part of a learning path? do we not want to include non-learning path content into a syllabus for a specific reason?

@pandasa123 @JRussellHuffman thoughts?

pandasa123: I think it should include everything but the `demo` / `example` ones

vabarbosa (changes requested):
see https://github.com/Qiskit/platypus/pull/855/files#r852286442

server/modules/syllabus/commands/create-syllabus/create-syllabus-dto.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  @MaxLength(64)
+  @IsString()
+  @IsNotEmpty()
+  readonly instructor!: string
+
+  @MaxLength(255)
+  @IsString()
+  @IsNotEmpty()
+  readonly location!: string
```
vabarbosa, line 22: @korgan00 it looks like these 3 fields (`name`, `instructor`, `location`) are required and the rest of the fields are optional. was there a discussion on what should be required and what should be optional? and if so can the UI be updated to indicate required fields.

in addition, when there is a server error (e.g., you don't fill out a required field), it is not surfaced in the UI and it looks like the `Publish` button does work (i.e., nothing is happening). there is no indication of a failure. 

checking the network tab you do see POST request but it fails and there is a failure response object.

@pandasa123 @JRussellHuffman do you have design and/or thoughts on what fields should be required and how it should be surfaced in the UI?

vabarbosa (changes requested):

server/modules/syllabus/commands/create-syllabus/create-syllabus-dto.ts
Tansito: @korgan00 and me had a discussion about what fields should be required or not. How we didn't know initially what we would can find or how the users will use the tool we decided the to put the minimum fields required as possible (initially `name`, `instructor` and `location`).

About the errors @vabarbosa you are totally right. Right now I'm not managing correctly the errors (mostly those one related with the validations) and is an improvement that I need to implement. What we talked @korgan00 and me was finish the main flow and after that try to fix these final things. In this case validate the input correctly and show to the user the feedback for those fields that need to be fixed.

Tansito: Btw, as a side note, the `instructor` is being filled with the user that creates the syllabus so it's automatically filled. The required parameter can be removed but we thought that it was a good idea to specify that that parameter must be always present in the Syllabus. We can have a discussion about it too, we were not totally sure about these decisions so really glad to hear more opinions.

pandasa123: The following fields should be required:

- Name
- Instructor

I feel like we should add a `* required` above the input field (or in the placeholder)

techtolentino: Just wanted to note that the `:required="true"` attributes _are_ included in the `<bx-input>` components, for example, shown on this [line here](https://github.com/Qiskit/platypus/pull/855/files#diff-e692348f820ccb24cddf18f987ae8ca9971c3d9f9245e95c27b78797915caf59R12)

Also, I believe this is _how_ that specific attr should be used, according to the Carbon web component's properties shown [here](https://web-components.carbondesignsystem.com/?path=/docs/components-input--form-item)... I tried adding a `required-validity-message` property locally, but nothing happens still. (Maybe we need to add our own validation check?)

Tansito: Let's remove the requirement for the `location` param then. But yeah, as @pandasa123 mentions above could be a good idea to add a `*` to the input to specify that they are required.

About the validation messages @techtolentino, I'm don't have ready a good way for you to check them yet. I have a conversation pending with @korgan00 and you about how we can do this but I need to finish first the refactor sprint. We can have a meeting for the next sprint if you see it ok!

frontend/vue/components/Syllabus/SyllabusView.vue
```diff
@@ -0,0 +1,142 @@
...
+    }
+  },
+  data () {
+    return {
+      sectionList: [] as Section[],
+    }
+  },
+  mounted () {
+    getCourseList().then((courses) => {
+      const learningPathCourses = courses.filter(course => course.type === 'learning-path')
```
pandasa123, line 71: We should have everything BUT docs

frontend/vue/components/Syllabus/SyllabusFormModule.vue
pandasa123: @techtolentino I think this needs to changed to `course.type != 'docs'`?

pandasa123 (changes requested):
Minor things to change the course filtering logic

server/modules/syllabus/commands/create-syllabus/create-syllabus-dto.ts
vabarbosa: the carbon component can check required fields, you need to call `checkValidity()` on the field: https://github.com/Qiskit/platypus/pull/928

Tansito: Oh yeah, sorry, when I said that I didn't have ready a way to check the validations I was thinking from backend to frontend. For only a frontend validation that's enough, totally.

eddybrando (approved):
After checking with @vabarbosa and @pandasa123, this is ready for shipping 🙂

PR #856, opened 2022-03-30
Fix db connection with interval wait

korgan00:
## Changes

Bugfix
Related with https://github.com/Qiskit/platypus/issues/847

## Implementation details

Interval waiting for the connection.

## How to read this PR

DB connection should work with a mongoose server and memory db


Tansito (approved):
Thanks for this fix @korgan00 !

PR #858, opened 2022-03-30
Get query params for pagination

### LINKED ISSUE #857: Syllabi end-point is not handling pagination

Syllabi end-point is not handling pagination params to manage the pagination.

`limit` and `page` query params should be editable by the front.

Tansito:
Fix #857

### DIFF
--- a/server/modules/syllabus/commands/find-syllabi/find-syllabi-controller.ts
+++ b/server/modules/syllabus/commands/find-syllabi/find-syllabi-controller.ts
@@ -13,10 +13,15 @@ export const FindSyllabiController = async (req: Request, res: Response, next: N
     return res.json(error)
   }
 
+  const { limit, page } = req.query
   const userId = req.user.id
 
   // Note: this object doesn't introduce external data so there is no need to validate it by now
-  const queryParams = new SyllabusQueryParamsHttpRequest({ owner: userId })
+  const queryParams = new SyllabusQueryParamsHttpRequest({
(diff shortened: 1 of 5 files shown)


server/modules/syllabus/commands/find-syllabi/find-syllabi-controller.ts
```diff
@@ -13,10 +13,15 @@ export const FindSyllabiController = async (req: Request, res: Response, next: N
...
   }
 
+  const { limit, page } = req.query
   const userId = req.user.id
 
   // Note: this object doesn't introduce external data so there is no need to validate it by now
-  const queryParams = new SyllabusQueryParamsHttpRequest({ owner: userId })
+  const queryParams = new SyllabusQueryParamsHttpRequest({
+    limit: (limit !== undefined ? parseInt(limit as string) : limit),
+    page: (page !== undefined ? parseInt(page as string) : page),
```
korgan00, line 25: I think you can handle more edge cases this way. If the `limit` is `NaN`, `undefined`, or `null` (I think null is not a case but is handled) then it gets unified into undefined. Also, you can just simplify with the `parseInt(limit as string)`, it will return a `NaN` in any edge case.

Non blocking

```suggestion
    limit: parseInt(limit as string) || undefined,
    page: parseInt(page as string) || undefined,
```

korgan00 (commented):
Hey this is great.

I have a question: are you using `page` anywhere?

korgan00 (approved):
LGTM

PR #883, opened 2022-04-08
patch mathigon studio

### LINKED ISSUE #799: User Account: Sync user progression in user account

Right now we are storing the progression in localhost and is not being connected with the user-account.

- [x] Verify if we are storing the progress right now

Connect that progressions with the user-account so once time the user logs in retrieves the progression of the user.

vabarbosa:
Fixes #799 

## Changes

patch the progress model in `@mathigon/studio` by changing the index of the `progresses` collection

## Implementation details

the progress model includes `userId` and `courseId` fields. however its index was referencing `user` and `course`.  as a result, when storing progress data it would fail. the field names in the index need to be updating to accurately coincide with the model field names.

---
if this PR gets merged the progresses collection index will need to be deleted so new one is created with this update

### DIFF
--- a/patches/@mathigon+studio+0.1.16.patch
+++ b/patches/@mathigon+studio+0.1.16.patch
@@ -314,6 +314,19 @@ index a28f2db..d1eb2e7 100755
    });
  
    app.get('/cron/cleanup', async (req, res, next) => {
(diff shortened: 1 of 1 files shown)


Tansito (approved):

Tansito:
good catch here 👀

PR #897, opened 2022-04-12
Improve content checks

frankharkins:
We have a few content checks in place, but they're not well explained / documented. This PR organises them in a way that should make it easier to add more checks, and adds a README to explain what each check does.

## Changes
- Moved content checks to `scripts/content_checks`.
- Changed some names to be more descriptive.
- Added a README.
- Removed `gitclean` filter for notebook metadata; I didn't realise that it runs on `git status` and it can really take a while. Instead, it's now a check like everything else, and you can run a different command to fix it.
- Removed the Makefile; probably best if there's only one 'correct' way to run the checks.


scripts/content_checks/goals.py
```diff
@@ -30,6 +30,7 @@ def check_file(filename: str, goal_names: List[str]) -> None:
             continue
         if filename.startswith('#'):
             print(f'Skipping: {filename}')
+            pass
         else:
-            print(f'Goals check: {filename}')
+            #print(f'Goals check: {filename}')
```
vabarbosa, line 35: should this just be removed instead of commented out?

vabarbosa (changes requested):
@frankharkins thank you! i just have one question before merging

scripts/content_checks/goals.py
frankharkins: Yes- thanks!

vabarbosa (approved):
thank you!

PR #904, opened 2022-04-13
Optimize SVGs

frankharkins:
## Changes

SVG is the preferred format for code outputs as they're crisper and smaller. The downside is that they use lots of lines of code that make for hard-to-read git diffs. In this PR I use scour to optimize the images and reduce them to a single line. I also add a script and a check to automatically optimize new changes and enforce optimized SVGs.

## Implementation details

Script uses [scour](https://github.com/scour-project/scour).

## How to read this PR

[First two commits](https://github.com/Qiskit/platypus/pull/904/commits/af20b7f7004061159d132e40081a9b50ac18e3b4) add the scripts and checks, the [third](https://github.com/Qiskit/platypus/pull/904/commits/6c7f7a08d22a179db60863e1c8c10c3287768fd7) optimizes all images in all notebooks.


vabarbosa (changes requested):
@frankharkins  this is awesome! thank you!

i am thinking it may be worthwhile to add the script as part of a pre-commit hook so the user is alerted of the need to optimize before the commit is even accepted (and not need to wait for build/test to run).

your thoughts?

related: https://github.com/Qiskit/platypus/issues/905

frankharkins:
@vabarbosa thanks! I've added a script to add the test to the pre-commit hook on `npm install`.

vabarbosa:
@frankharkins thank you for this! 

i updated this a bit to use the [git hook setup being worked on](https://github.com/Qiskit/platypus/pull/941). this allows the [script to only run on the files being committed at that time](https://github.com/Qiskit/platypus/blob/c402efae0b4368b562f51b61e4f2c2dab18a31fd/package.json#L39-L45) that way we do not necessarily need to have it run across all file that had been updated/being committed.

for this PR i only updated the metadata and svg checks, but in subsequent PR we can update the other checks.

your feedback is welcomed so let me know if you have any questions or you see any issues with the changes i made.

frankharkins:
Look good, thank you!

vabarbosa (approved):

Issue #837, opened 2022-03-22
User account: Syllabus template and user account


techtolentino:
### Steps to reproduce the problem

1. Log in at `/account`
2. Go to `/syllabus/abc-123` or `/syllabus/create`
3. See that if a user is logged in, they cannot see the side bar

### Designs
[EDIT: Updated link]
Designs for this can be found [here](https://www.figma.com/file/f1QMqiXSgt2FCt0itvsA24/676-Professor-Syllabus?node-id=814%3A8646)

### What is the current behavior?

The user account sidebar doesn't show if a user is logged in

### What is the expected behavior?

- If a user is logged in, the user account sidebar should show
- a user can only see `syllabus/create` if they're logged in
- a user does not need to be logged in to see `syllabus/abc-123`, but if they are, the designs say that the sidebar should have `Classroom` highlighted, and the sidebar should show

techtolentino:
tagging @JRussellHuffman and @gracelindsell just for visibility.

@eddybrando - please let me know if I should schedule a call so we can discuss the UX here... 

thanks

JRussellHuffman:
Can someone post a URL I can follow to see the bug? From context, it seems like it should be learn.qiskit.org/syllabus/create or learn.qiskit.org/account/syllabus/create, but neither work for me. 

Happy to talk through the UX, but so far this sounds like a bug, but I probably don't have all the context yet.

PR #923, opened 2022-04-20
Refactor `/account` routes

### LINKED ISSUE #867: Reconsider routing for `/account` page
For scalability purposes, we could reconsider the way we handle routes, in regards to the user `/account` page.

Currently, all views under the user account are based on an _active_ section `hash`, as seen [here](https://github.com/Qiskit/platypus/blob/main/frontend/vue/components/UserAccount/UserAccountLayout.vue#L44-L52). However, with this approach, we run into an issue with scaling — for example, trying to implement something like `account > classroom > syllabus > create`... or `account > classroom > syllabus > edit`.

This issue is to propose changing this strategy from using the _sectionId_ approach, in favor of a URL-based approach(?) so that we have routes such as 

- `/account/mylearning`
- `/account/classroom`
- `/account/classroom/syllabus/create`
- `/account/classroom/syllabus/edit`

techtolentino:
## Changes

Fixes #867 

## Implementation details
- add new routes for `/account/classroom`, `/account/privacy`
- replaced `hash` based approach, with `pathname` for `activeSection` matching
- replaced `AppLink` w/ `BasicLink` for account sidebar

## How to read this PR
- review `getAccountData` in `app.ts` to see if this is the right approach
- review UI (or screen grab below)

## Screenshots

https://user-images.githubusercontent.com/6276074/164274419-1c464477-2e55-454c-a4ad-c47dc164780c.mov

## Notes
- I'm making this PR into the `feature/professor-syllabus` branch, but please let me know if this should come _after_ that branch is merged. In other words, should this be part of a later Syllabus 1.1 milestone?
- This PR does **not** handle the `/syllabus` related routes...  I think that needs to be considered in separate refactor, along with [this issue](https://github.com/Qiskit/platypus/issues/837)

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -98,8 +98,8 @@ const getAccountData = async (req: Request, res: Response) => {
   const privacyPolicyMD = loadLocaleRawFile('privacy-policy.md', lang)
 
   const userMockData = {
-    firstName: req.user?.firstName,
-    lastName: req.user?.lastName
+    firstName: req.user.firstName,
+    lastName: req.user.lastName
   }
 
   const progressData = await getUserProgressData(req)
--- a/server/app.ts
+++ b/server/app.ts
@@ -172,6 +172,66 @@ const start = () => {
         translationsJSON
       })
     })
+    .get('/account/classroom', async (req, res) => {
+      if (!req.user) {
+        return res.redirect('/signin')
+      }
+      if (req.user && !req.user.acceptedPolicies) {
+        return res.redirect('/eula')
+      }
+
+      const lang = req.locale.id || 'en'
+      const translationsJSON = JSON.stringify(TRANSLATIONS[lang] || {})
+
+      const privacyPolicyMD = loadLocaleRawFile('privacy-policy.md', lang)
+
+      const userMockData = {
+        firstName: req.user?.firstName,
+        lastName: req.user?.lastName
+      }
+
+      const progressData = await getUserProgressData(req)
+
+      res.render('userAccount', {
+        config: CONFIG,
+        userData: userMockData,
+        progressJSON: JSON.stringify(progressData),
+        lang,
+        privacyPolicyMD,
+        translationsJSON
+      })
+    })
+
+    .get('/account/privacy', async (req, res) => {
+      if (!req.user) {
+        return res.redirect('/signin')
+      }
+      if (req.user && !req.user.acceptedPolicies) {
+        return res.redirect('/eula')
+      }
+
+      const lang = req.locale.id || 'en'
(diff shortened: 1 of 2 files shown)


server/app.ts
```diff
@@ -84,6 +84,36 @@ const getUserProgressData = async (req: Request) => {
...
+    return res.redirect('/eula')
+  }
+
+  const lang = req.locale.id || 'en'
+  const translationsJSON = JSON.stringify(TRANSLATIONS[lang] || {})
+
+  const privacyPolicyMD = loadLocaleRawFile('privacy-policy.md', lang)
+
+  const userMockData = {
+    firstName: req.user?.firstName,
```
Tansito, line 101: How we are checking if the `user` exists in the first line of the method I think we can assume that the user exists at this point and remove the `?` after `req.user`.

Tansito (changes requested):

Tansito:
Thanks @techtolentino ! I just commented a little change that I saw but apart from that for me it's ok, do you think that for syllabus the routes will need a refactor? We can try to do it for Syllabus 1.1.

server/app.ts
```diff
@@ -84,6 +84,36 @@ const getUserProgressData = async (req: Request) => {
   return progress
 }
 
+const getAccountData = async (req: Request, res: Response) => {
+  if (!req.user) {
+    return res.redirect('/signin')
+  }
+  if (req.user && !req.user.acceptedPolicies) {
+    return res.redirect('/eula')
+  }
```
vabarbosa, line 93: along the lines of what @Tansito mentioned, at this point we already know we have `req.user` (checked in statement above this one) so no need to check it here

```suggestion
  if (!req.user.acceptedPolicies) {
    return res.redirect('/eula')
  }
```

server/app.ts
```diff
@@ -84,6 +84,36 @@ const getUserProgressData = async (req: Request) => {
...
+
+  const lang = req.locale.id || 'en'
+  const translationsJSON = JSON.stringify(TRANSLATIONS[lang] || {})
+
+  const privacyPolicyMD = loadLocaleRawFile('privacy-policy.md', lang)
+
+  const userMockData = {
+    firstName: req.user.firstName,
+    lastName: req.user.lastName
+  }
```
vabarbosa, line 103: does `userMockData` still make sense here as a variable name? maybe `loggedInUser` or just `userData`

vabarbosa (approved):
this is great @techtolentino! just had a couple very minor suggestions

server/app.ts
vabarbosa: thanks looks good 👍🏽

PR #924, opened 2022-04-21
Escape curly brackets twice in inline equations.

frankharkins:
## Changes

Latex such as `\{0, 1\}` should render as {0,1}, i.e. show the curly brackets, but at the moment they don't display on the site. Escaping twice brings them back, so I added some logic in the converter to escape these again.


vabarbosa (approved):
@frankharkins nice... thank you!

PR #925, opened 2022-04-21
Fixed as required only the name and instructor

Tansito:
## Changes

From #855 we decided to left only as required specifically the params `name` and `instructor`.

## Implementation details

All the _Syllabus_ parameters excepted `name` and `instructor` are configured as **optional**.


vabarbosa (approved):
@Tansito thank you!

PR #928, opened 2022-04-21
syllabus review feedback

vabarbosa:
## Changes

updates based on some review feedback
- https://github.com/Qiskit/platypus/pull/855#discussion_r852199630
- https://github.com/Qiskit/platypus/pull/855#discussion_r853736675

**Chapters** title is now hidden when it contains no content.

form is no longer submitted with empty required fields. users are provided with a validity message.

## How to read this PR

- review updates to syllabus form and view files
- create/edit syllabus with & without **Name** and/or **Instructor**

## Screenshots

image

PR #929, opened 2022-04-22
Check input pristine status

Tansito:
## Changes

From #928 this PR improves the UX to just validate those inputs that went modified before show an error.

## How to read this PR

- modify `name` or `instructor` and check if the validations show an error in other inputs that were not modified.

## Screenshots

Screenshot 2022-04-22 at 12 26 18

How the user didn't reach to the instructor field that input is `pristine` and is not validated.


vabarbosa (approved):
@Tansito thank you!

Issue #397, opened 2021-11-18
"Name Error" in code cell


JRussellHuffman:
### Steps to reproduce the problem

1. Go to https://learn.qiskit.org/course/ch-demos/level-5-proving-the-uniqueness-of-quantum-variables
2. Run the first code cell
3. See error

### What is the current behavior?
The code throws an error "name 'hello_quantum' is not defined"

### What is the expected behavior?

I assume the first code cell on a page should always run. Further down the page, there may be dependencies on prior code cells, but the first cell should run correctly.

### Screenshots (optional but very helpful)

<img width="743" alt="Screen Shot 2021-11-18 at 10 18 32 AM" src="https://user-images.githubusercontent.com/7633881/142454398-58dd246d-0775-43ed-9f69-90abef0ae750.png">

### Environment (optional but very helpful)
 - Device: macbook pro
 - OS: osx
 - Browser: firefox

issue is probably on all environments

### Additional context (optional)

Add any other context that could help debug the problem.

bopardikarsoham:
This is because the hello_quantum module has not been imported from qiskit_textbook.games

![image](https://user-images.githubusercontent.com/77266161/165014766-bd45c616-6754-4ed2-9e88-a2a8f9fb6a1b.png)

Simply by adding `from qiskit_textbook.games import hello_quantum` fixes the issue

bopardikarsoham:
But, I also found out that in https://github.com/Qiskit/platypus/blob/main/notebooks/ch-ex/hello-qiskit.ipynb all the levels have been merged into one jupyter notebook. But, if we look at the textbook  https://learn.qiskit.org/course/ch-demos/level-5-proving-the-uniqueness-of-quantum-variables Level 5 is created like a separate chapter. Why are they different?

PR #937, opened 2022-04-25
Fixed Issue-#397

bopardikarsoham:
Fixed "Name Error" as per the discussion in issue #397

### DIFF
--- a/notebooks/ch-ex/hello-qiskit.ipynb
+++ b/notebooks/ch-ex/hello-qiskit.ipynb
@@ -1932,6 +1932,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "from qiskit_textbook.games import hello_quantum\n",
     "initialize = [['x', '0'],['cx', '1']]\n",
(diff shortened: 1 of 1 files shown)


Tansito:
I add @frankharkins to the conversation!

frankharkins (changes requested):
Looks like `ch-ex/hello-qiskit.ipynb` is incorrectly being split into three different pages because there's more than one top-level heading. I think the correct fix for #397 is to make all except the top heading second-level, then all code cells will be on the same page and we won't get an error. We can maybe add a check for this too.

vabarbosa:
> Looks like `ch-ex/hello-qiskit.ipynb` is incorrectly being split into three different pages because there's more than one top-level heading. I think the correct fix for #397 is to make all except the top heading second-level, then all code cells will be on the same page and we won't get an error. We can maybe add a check for this too.

FYI the heading issue with `ch-ex/hello-qiskit.ipynb` has been addressed as part of #954

PR #948, opened 2022-04-27
Update Syllabus template with conditional account menu

### LINKED ISSUE #837: User account: Syllabus template and user account

techtolentino:
## Changes

Fixes #837 
Co-authored by @vabarbosa 

## Implementation details

- abstracted `AccountMenu` component from `UserAccountLayout`
- used `AccountMenu` component in `SyllabusLayout`
- pass user data into `Syllabus.pug` template, down to components
- used `req.session` to temporarily add redirect url back to syllabus

## How to read this PR
- review `AccountMenu` component
- review use of `AccountMenu` in both syllabus and user account layouts
- review `redirectTo` approach in `app.ts`

## Screenshots

https://user-images.githubusercontent.com/6276074/165812854-39aab216-de76-43b6-8d2d-f350a54394e4.mov


translations/strings.yaml
```diff
@@ -14,6 +14,7 @@ Copied!: ''
 Copy: ''
 Create a customizable syllabus and share it with your students.: ''
 Create a syllabus: ''
+create an account: ''
```
Tansito, line 17: It seems that we are starting always in the translations with a capital letter, should we change it to `Create an account`?

server/app.ts
```diff
@@ -85,6 +83,15 @@ const getUserProgressData = async (req: Request) => {
 }
 
 const getAccountData = async (req: Request, res: Response) => {
+  // @ts-ignore
```
Tansito, line 86: I created the PR #966 to try to remove these `ts-ignore's`. Let me know that you think.

Tansito (changes requested):

translations/strings.yaml
techtolentino: @Tansito - thanks for the review 👋 

I think we can leave this one as a lowercase, since this particular string starts in the middle of a sentence. Here's the design: 

Screen Shot 2022-05-02 at 5 23 59 AM

server/app.ts
techtolentino: Awesome, thank you... I'll have a look 👏

Tansito (approved):
LGTM! Thanks @vabarbosa @techtolentino really nice! 💯

Issue #951, opened 2022-04-29
build failure with `simple-git-hooks`


vabarbosa:
### Steps to reproduce the problem

1. trigger a release/deployment to production
2. view result of the action: https://github.com/Qiskit/platypus/runs/6190013441?check_suite_focus=true
3. the deployment fails trying to setup `simple-git-hooks`

### What is the current behavior?

releases are failing

### What is the expected behavior?

successful release/deploy

### Screenshots (optional but very helpful)

**failed build**
https://github.com/Qiskit/platypus/runs/6190013441?check_suite_focus=true
<img width="509" alt="image" src="https://user-images.githubusercontent.com/13156555/165972545-a7be404b-e67f-4c7b-8527-b19ab551cb36.png">

PR #952, opened 2022-04-29
build error with `simple-git-hooks`

vabarbosa:
## Changes

Fixes https://github.com/Qiskit/platypus/issues/951

disable setting up git hooks when deployment. it is not needed and is causing build failures.

## Implementation details

overwrite the `prepare` script in `package.json` before running `npm ci` so git hooks setup is not attempted

## How to read this PR

1. review update to `release.yml`
2. see previously failed build due to `simple-git-hooks`: https://github.com/Qiskit/platypus/runs/6190013441?check_suite_focus=true
3. see successful build after removing `simple-git-hooks` setup: https://github.com/Qiskit/platypus/runs/6222812881?check_suite_focus=true

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,6 +27,9 @@ jobs:
           node-version: "16"
           cache: "npm"
 
+      # overwrite "prepare" script in deployment since git hooks are not needed
+      - run: npm set-script prepare ""
+
       - run: npm ci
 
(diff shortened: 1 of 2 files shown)


techtolentino (approved):
Discussed this with @vabarbosa 

checked the before (broken build) and the successful one after...

LGTM 👏

PR #965, opened 2022-05-02
Logger system

### LINKED ISSUE #964: Improve logs management
Improve the logs system with the purpose to improve traceability and processing.

Tansito:
## Changes

Fix #964 

## Implementation details

- Added `pino` as the new library to manage logs. This library is very standard in node's ecosystem and it's the same library used in IBM Cloud.
- I tried to copy the same configuration as we have in IBM Cloud. The idea is try to reproduce the same environment.
- `pino-prettier` will make easier to read the logs for when we are in a development environment.
- `pino-http` is attached to MathigonApp and it will log automatically the accesses to our system externally.

## How to read this RP

- Two new files: `logger` and `logger-express`. Both are the instances to manage the logs. Only the first should be used manually, the second one is for MathigonApp.
- `transports.ts` is the file where to configure the plugins that will be added to `pino`. By now we only have `pino-prettier` but very useful if in the future we need to store the logs in a persistent storage, to put an example.

## Screenshots

Screenshot 2022-05-02 at 10 34 53


server/libs/logger/logger-express.ts
```diff
@@ -0,0 +1,11 @@
+import { stdTimeFunctions } from 'pino'
+import { pinoHttp } from 'pino-http'
+
+const isProd = process.env.NODE_ENV === 'production'
```
vabarbosa, line 4: do we know if this is working as intended? i believe there was an issue with this env variable for the cleanAndPopulate?

server/libs/logger/logger.ts
```diff
@@ -1,2 +1,30 @@
-// eslint-disable-next-line no-console
-export const logger = () => console.log('TODO: IMPLEMENT LOGGER')
+import { pino, stdTimeFunctions } from 'pino'
+import type { Logger, LoggerOptions } from 'pino'
+
+import { pinoPrettyTransport } from './transports'
+
+const isProd = process.env.NODE_ENV === 'production'
```
vabarbosa, line 6: we keep making this check throughout. would be useful/beneficial to just make it once in particular file and import as needed?

vabarbosa (changes requested):
@Tansito this is great. thank you! just had a couple minor points.

in addition, perhaps you can add a `pre-commit` hook that checks and make sure someone doesn't add/commit `console.log`. if that will not be a straightforward script, we can just open an issue to work on it at a later date.

server/libs/logger/logger-express.ts
Tansito: Yes! We are going to need to resolve #975 but this is the logic that we are going to need to follow. I will try to solve this today.

server/libs/logger/logger.ts
Tansito: I thought the same but how both loggers can have different configuration I decided to separate it. I will upload a change merging both configurations and we can decide 👍

Tansito: Now I read correctly your comment. Yeah, totally right, I will check if Mathigon provides us with that functionality if not I can create some kind of utility.

server/libs/logger/logger-express.ts
Tansito: Opened PR #984

PR #966, opened 2022-05-02
Added redirectTo property to the SessionData

Tansito:
## Changes

From #948 

## Implementation details

- Created a new `d.ts` file for `express-session` module that extends the current `SessionData`.
- Configured the `tsconfig` to include this file
- Added `--files` option in **ts-node**. I would like to give more context here because the investigation was a bit difficult:
  - [Issue](https://github.com/TypeStrong/ts-node/issues/782) about including custom types in **ts-node**
  - [Documentation](https://github.com/TypeStrong/ts-node#help-my-types-are-missing) from **ts-node**
  - Seems that ts-node is having a hard time handling custom types from users. Initially they have a documentation where they explain a configuration step by step but as you can read in the issue it gives more headaches than solutions (I couldn't it make it work either). So I opted for the `--files` option that enables the standard **tsc** behaviour and takes into account the `includes` folder. This way I could make it work.
- I only added the `--files` option in the development command because is where we check the types.

## How to read this PR

- Added support for types extension: **express-session** in this case.
- Removed the unneeded `ts-ignore`

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "watch:studio": "mgon-build --assets --watch",
     "watch": "npm-run-all --parallel watch:vue watch:studio",
     "start": "ts-node --transpile-only -I public -P server/tsconfig.json server/app.ts",
-    "start-dev": "nodemon --watch 'server/**/*.ts' --exec 'ts-node -I public -P server/tsconfig.json server/app.ts'",
+    "start-dev": "nodemon --watch 'server/**/*.ts' --exec 'ts-node --files -I public -P server/tsconfig.json server/app.ts'",
     "dev": "npm-run-all --parallel watch start-dev",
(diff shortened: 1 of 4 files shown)


techtolentino (approved):
@Tansito - this looks good to me. I read through your implementation notes and followed through links, thanks for including the summary.

I think this looks good. 

Will we be able to extend the `d.ts` file in further use cases if we need?

vabarbosa (approved):
thank you!

Tansito:
> @Tansito - this looks good to me. I read through your implementation notes and followed through links, thanks for including the summary.
> 
> I think this looks good.
> 
> Will we be able to extend the `d.ts` file in further use cases if we need?

Yeah, the idea is that in the future if we need to extend the type of some module or even this one we can follow the same logic.

PR #984, opened 2022-05-04
Setup NODE_ENV

### LINKED ISSUE #975: Environment configuration
### What is the current behaviour?

If we want to deliver production code the env var of the project is not correctly configured.

### What is the expected behaviour?

Configure correctly the environment variables for platypus.

Tansito:
## Changes

Fix #975 

## Implementation details

I just followed the steps described in [App Engine documentation](https://cloud.google.com/appengine/docs/standard/nodejs/config/appref#environment_variables).

It seems very simple but how at the end of the day we only use App Engine in production, this should be more than enough.

## How to read this PR

I just added the `env_variables` property to the `app.yaml` and this should be more than enough to configure the environment. We can check this functionality with #965 .

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -17,3 +17,6 @@ liveness_check:
 
 readiness_check:
   path: "/health"
+
+env_variables:
+  NODE_ENV: "production"


abdonrd (approved):

vabarbosa (approved):
thank you! 🤞🏽

vabarbosa:
@Tansito once this is deployed, do we have a way to confirm test before we re-enable the `cleanAndPopulate`?

Tansito:
We can for example confirm this PR once time we merged #965

vabarbosa:
> We can for example confirm this PR once time we merged #965

oh ok. do we have access to see the logs in AppEngine?

Issue #976, opened 2022-05-03
Progress for unauthenticated users is being stored


Tansito:
### What is the current behaviour?

Right now we are storing in the database the progression for not authenticated users what it generates a lo of not needed entries in the database.

### What is the expected behaviour?

- Stop storing the progress for not authenticated users in the database.
- Clean the database with those entries without a relation with our current users

PR #988, opened 2022-05-06
Avoid storing the progress in DB for not authenticated users

Tansito:
## Changes

Related with #976 

## Implementation details

- Mathigon generates a [tmpUser](https://github.com/mathigon/studio/blob/main/server/app.ts#L258) when there is no user authenticated.
- In the POST end-point to store the progression what I did was accept only authenticated users through a patch.
- And only authenticated users will be able to get data from database.
- Those are the end-points that I could identify as problematics in the flow.

## How to read this PR

- There is a patch where I'm adding a check in case there is no user and we are trying to store the progress.
- I removed the tmpUser from our method to retrieve the user's progress

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -70,7 +70,7 @@ const getCourseData = async function (req: Request) {
 }
 
 const getUserProgressData = async (req: Request) => {
-  const userId = req.user?.id || req.tmpUser || null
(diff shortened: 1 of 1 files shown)


vabarbosa (approved):
thank you!

PR #994, opened 2022-05-10
Fix compatibility dependencies

### LINKED ISSUE #992: Fix dependencies error at installation

PR updates are failing due to a problem between dependencies versions, see: #989 , #990 , #991 ...

Not fail at installation.

Tansito:
## Changes

Fix #992 

## Implementation details

To fix the compatibility problemas between dependencies I needed to take tough decisions that I would like to comment:

- `vue-router`: fixed version to `4.0.12`. This was needed because for greater versions it needs **vue: ^3.2.0** and we have some dependency that it's not compatible (I couldn't find which one, we are going to need more time to investigate it). Due to the changes in the runtime it seems that components above _3.2.0_ are not compatible with components below _3.2.0_ so we are going to need check this (more information in its [CHANGELOG](https://github.com/vuejs/core/blob/main/CHANGELOG.md#compatibility-notes)).
-  `node-sass`: reduced the version to `6`. This was required due to the `sass-loader` version. I tried to upgrade the `sass-loader` version but we have some kind of compatibility problem at build time. This problem will require more investigation (and could be related with Mathigon version that we are pending to upload).
- `@carbon/icons-vue`: limited the version to `1.48.X`. This is curious because it seems that in **1.49** they removed a component that we are using. Another thing to investigate.
Screenshot 2022-05-10 at 12 33 55

These are the changes that I needed to apply to solve the problems with the compatibility dependencies. I will create the issues related with the things that I found.


vabarbosa (approved):
@Tansito thank you for getting to the bottom of this! greatly appreciated.

PR #996, opened 2022-05-11
New mock data generation process

### LINKED ISSUE #977: Fix the database population script for local development and staging
### What is the current behaviour?

We had a problem in the past with the database population script and we needed to disable it. 

### What is the expected behaviour?

- It depends on #975 
- Review the current script to not remove entries just only create entities if they don't exist

Tansito:
## Changes

Fix #977

## Implementation details

- Removed the deletion of the entries. Now the mock data will be created in case the entry doesn't exist.

## How to read this PR

- Added the script to the `app.ts`
- Modified the population process in `populate-database.ts`


vabarbosa (approved):
LGTM!
@Tansito thanks

PR #1002, opened 2022-05-13
Automatically update mega menu from notebooks/toc.yaml

### LINKED ISSUE #955: Megamenu course links should not be hard coded
the megamenu containing links to all courses currently maintains [a hard-coded list of courses and links](https://github.com/Qiskit/platypus/blob/main/frontend/vue/components/constants/megaMenuLinks.ts).
this should obtained programmatically so if when course get added/changed the megamenu stays current

related https://github.com/Qiskit/web-components/issues/74

frankharkins:
## Changes

Mega menu now gets links automatically from `notebooks/toc.yaml` instead of from hard-coded file (fixes #955).

## Implementation details

- `vue.config.js` reads the `toc.yaml` on build, parses it, and injects it into `megaMenuLinks.ts`
- `megaMenuLinks.ts` then converts this to the form that the mega menu uses (this step could potentially be moved to `vue.config.js` to happen on build).
- I added a `type` to the old textbook chapters in the TOC, which we can use to decide which pages show up on the menu

## How to read this PR

1. Read `frontend/vue/vue.config.js`
2. Read `frontend/vue/components/constants/megaMenuLinks.ts`

## To do

- [ ] Potentially add a better algorithm for organising the columns more evenly


vabarbosa (changes requested):
@frankharkins thanks for this. 

the [megamenu is a shared component ](https://github.com/qiskit-community/qiskit-vue/tree/main/src/components/mega-menu-dropdown)between platypus and qiskit.org. unfortunately, this update does resolve it for playtpus but still have issue on the [qiskit.org side](https://github.com/Qiskit/qiskit.org/blob/main/constants/megaMenuLinks.ts) (textbook-beta home page). would it be possible to update the shared component instead to pull the content from a specified endpoint (e.g., https://learn.qiskit.org/courseList). or perhaps you can think of a better approach so all places the megamenu is used will be consistent and update automatically.

CC @eddybrando because i believe he may have some thoughts around this and i had just brought this up to him earlier before seeing your PR.

let me know if i am not making any sense or you have any questions. thanks!

PR #1020, opened 2022-05-17
Korean translation added

0sophy1:
Korean translation on-boarded!

## Changes
Added Korean translation and TL onboarded


0sophy1:
It's so strange that I only changed LOCALIZATION_CONTRIBUTORS's file

techtolentino (approved):
@0sophy1 - this isn't your fault, it's an internal problem where we've met our storage quota for branch deploys... 

I think your PR is fine.

Approved!

🚢

PR #1033, opened 2022-05-18
Added German translation

0sophy1:
Added @TMuehge and German

He already joined gitlocalize


veenaiyuri:
@0sophy1 Hi Sophy, thanks for your support. It is important that @TMuehge opens a PR himself to add his name to the LOCALIZATION_CONTRIBUTORS list.

PR #1039, opened 2022-05-19
Nano: fix notebooks on commit

frankharkins:
## Changes

Instead of raising an error, commits now silently fix the notebook SVGs and metadata on commit.

## Reason

Without this change, Nano will run checks and block commits on any staged notebooks. Running `npm run test:nb:fix` only fixes notebooks in `scripts/content_checks/notebook_paths.txt`, meaning contributors have to modify `notebook_paths.txt`, or manually edit the metadata.


techtolentino (commented):
I like what this is doing, @frankharkins 

Do you have thoughts on how I could test this locally, to see it silently fixing?

frankharkins:
You could edit a notebook with Jupyter notebook, save it, then view the diff to see the changed metadata. Then stage, commit, and check the diffs between the commits.

For the SVGs you'll need to set your output type to SVG by running `%config InlineBackend.figure_formats = ['svg']` in the notebook (or [change it permanently](https://learn.qiskit.org/course/ch-prerequisites/environment-setup-guide-to-work-with-qiskit-textbook)), and follow the same process.

techtolentino (approved):
Thanks for this @frankharkins 

I was able to get this working locally.

**Note:** I had to use `pip3` for installing the dependencies... and it worked after

PR #1051, opened 2022-05-24
Add rich text editor to syllabus form

### LINKED ISSUE #1001: Investigate / implement Vue rich text editor
As a Professor creating a Textbook syllabus, I need the ability to add custom links and have proper formatting when adding syllabus content. 

This story is to investigate (and maybe prototype) a simple implementation of a Vue (or JS) rich text editor plugin or package. 

Minimal criteria
- [x] share findings on what RTE packages/plugins would work
- [x] allow a professor to include custom links in the content section of the syllabus block

**Syllabus form textarea that needs the additional functionality**
Screen Shot 2022-05-12 at 11 31 09 AM

techtolentino:
## Changes
Fixes #1001 

## Implementation details
- added [CKEditor](https://ckeditor.com/docs/ckeditor5/latest/installation/getting-started/frameworks/vuejs-v3.html) for rich-text editing
- added necessary CKEditor packages
- used CKEditor component for the syllabus content description field
- updated template to render html (output of rte)

## How to review PR
- pull down branch to your local machine, and `npm ci`
- make sure you can auth in using the Google auth approach
- go edit your Classroom > Syllabus
- Publish

## Screenshots

https://user-images.githubusercontent.com/6276074/170350535-bfdfa944-bafe-4cfe-9d3a-2b9b7db61052.mov


vabarbosa (approved):
@techtolentino this is awesome! great work. i just have a couple comments for possible consideration. none of these i think should be blockers so i am approving. depending on the difficulty to resolve, these could be addressed in separate issues later or as part of this PR (i let you make the decision).

- change default textarea height greater than 1 line (maybe 10 lines?)

    **current**
    image

    **with 10 lines**
    image

- when adding a link if you dont include protocol (e.g., `http`) the link is assumed to be relative

    image

- _italics_ format is not working for me

    **setting italics**
    image

    **what is displayed**
    image

techtolentino:
@vabarbosa - great feedback. We'll still need to figure out the URL part -- you're right, using something like `example.com` will parse as `learn.qiskit.org/course/path/example.com`... which we don't want.

We'll have to let users know (@pandasa123), that in the meanwhile, we'll need to use `https` / `http` when adding URLs.

In regards to the _italics_ and `min-height` for the editor, I updated the styles:


https://user-images.githubusercontent.com/6276074/170350535-bfdfa944-bafe-4cfe-9d3a-2b9b7db61052.mov

vabarbosa:
> @vabarbosa - great feedback. We'll still need to figure out the URL part -- you're right, using something like `example.com` will parse as `learn.qiskit.org/course/path/example.com`... which we don't want.
> 
> We'll have to let users know (@pandasa123), that in the meanwhile, we'll need to use `https` / `http` when adding URLs.
> 

@techtolentino thanks for the quick turnaround. you probably already saw this but regarding relative links. it looks like there is a CKeditor setting (`config.link.defaultProtocol`) to set a default protocol when none is specified: https://ckeditor.com/docs/ckeditor5/latest/features/link.html#adding-default-link-protocol-to-external-links

techtolentino:
@vabarbosa - I did not see that... this is good to know!

vabarbosa (approved):
thank you @techtolentino 🚢

vabarbosa:
> @vabarbosa - I did not see that... this is good to know!
> 
> I like your suggestions for both adding the default protocol, and for having links open in a new tab... great UX thinking.
> 
> Still, I think there's a couple of things to figure out here since the docs are targeting an `id="editor"`... but we'll be having more than one editor per form, potentially.
> 
> What if we merge as is, and enhance/improve along the way?

👍🏽 those can definitely be enhancements

Issue #416, opened 2021-11-04
Latex doesn't render in header


JRussellHuffman:

JRussellHuffman:
Possibly related to https://github.com/Qiskit/platypus/issues/980

PR #1080, opened 2022-05-31
Fix latex in headers

techtolentino:
## Changes

Fixes https://github.com/Qiskit/platypus/issues/416

## Implementation details

- used Markdown syntax instead of HTML, in markdown cells
- this helped the latex render

## Notes
@frankharkins - I could use your guidance here. Please take a look at the video below. Notice that the headers are fixed when I change an `<h3>` to `###`... however, these items then get added to the side navigation. 

Is that something we want? Also, the numbering will need to be updated in the side nav.

All in all, it seems like the information hierarchy of this page, specifically, could use some guidance.

Your feedback is appreciated 🙏 

## Screenshots

https://user-images.githubusercontent.com/6276074/171299143-dd01c6db-940c-49d9-b486-17aea4673834.mov

### DIFF
--- a/notebooks/ch-labs/Lab03_QuantumMeasurement.ipynb
+++ b/notebooks/ch-labs/Lab03_QuantumMeasurement.ipynb
@@ -66,9 +66,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


Tansito (approved):
Really good catch here @techtolentino 💯

frankharkins (approved):
Yes I think these should technically have been showing in the side nav; looks great thank you!

PR #1159, opened 2022-06-16
SymbolicNotation Fix

korgan00:
## Changes

Fix https://github.com/Qiskit/platypus/issues/1158

This is hard to test because it happened in the chapter that is not merged and it doesn't happen on the widgets page.


frankharkins (approved):
I approve

PR #967, opened 2022-05-02
Pull in changes from qiskit-community/qiskit-textbook

frankharkins:
## Changes

Fixes #921 & #427.

There have been changes to the notebooks in both this repo, and [qiskit-community/qiskit-textbook](https://github.com/qiskit-community/qiskit-textbook). This PR attempts to incorporate the positive changes from [qiskit-community/qiskit-textbook](https://github.com/qiskit-community/qiskit-textbook) without overwriting changes made in this repo.

## Implementation details

I copied the notebooks over from [qiskit-community/qiskit-textbook](https://github.com/qiskit-community/qiskit-textbook) and went through each notebook with git patch to try and keep the correct changes. I also added some new notebooks (resolves #865, #784, #221, #131). 

## How to read this PR

This diff is huge so maybe best thing is to spend most time testing the preview? I haven't properly tested this yet, so it's likely some things are broken, if you have time, please test and let me know!

PR #1160, opened 2022-06-16
Increase default space size for release gh action

Tansito:
## Changes

From #967 we needed to increase the default size for the tests and it seems that [the error size](https://github.com/Qiskit/platypus/runs/6920018084?check_suite_focus=true#step:11:688) that we are having in the release gh-action it's related with the same cause.

I moved the change applied by Va in `test.yml` to `release.yml`. Let's see if with that we solve the problem.


techtolentino (approved):
Thanks @Tansito 🤞 

🚢

PR #1209, opened 2022-06-22
Created the entity to store the custom fields

### LINKED ISSUE #1197: [Q Hardware] Table creation to manage user custom fields
### Feature description

By #1196 we are going to need to store the API Key for the user. How Mathigon doesn't provide an easy way to modify its user  we are going to need to create a new table with a relationship with the mathigon user to store custom fields.

Tansito:
Fix #1197 

## Change

Created the table `qiskit-user` where it will live all our custom fields required by us. By now we only have `apiToken` but we could have whatever we need.


techtolentino (approved):
Nicely done, @Tansito 

🚢

Issue #1214, opened 2022-06-22
Bug: Button Tooltip position causing overflow in Utility Panel


techtolentino:
### Steps to reproduce the problem

1. Go to Scratchpad Panel (will need to pull down the `feature/scratchpad_1.0` branch)
2. Scroll left/right within the right side Scratchpad panel
3. See overflow issue

### What is the current behavior?

Slight overflow issue is being caused by the Reset `bx-tooltip-icon`

### What is the expected behavior?

The tooltip shouldn't cause the overflow issue

### Screenshots

https://user-images.githubusercontent.com/6276074/175141000-77da30c8-d043-4e79-873a-cd861c162ce2.mov

techtolentino:
Fixed

PR #1215, opened 2022-06-22
Update attribute typo and `alignment` value

techtolentino:
## Changes

Fixes https://github.com/Qiskit/platypus/issues/1214

## Implementation details

- corrected spelling of `alignment` attribute/property
- updated alignment value to fix Scratchpad panel overflow issue

## Screenshots of before/after

https://user-images.githubusercontent.com/6276074/175141475-f04cacc6-142e-4b00-8e09-3729001db955.mov


Tansito (approved):
Reviewing the feature maybe it's interesting that the code-cell adapts its width to the container too to avoid the scroll but the fix at least looks good to me 👍

techtolentino:
@Tansito -- good point... we need to keep this in mind going forward; but I agree 💯

PR #1221, opened 2022-06-24
Use Vale prose linter to catch writing errors

frankharkins:
## Changes

Use [Vale](https://vale.sh/) prose linter to ensure writing quality across notebooks. Most important functionality is the spell checker, but it also comes with other useful features.

## Implementation details

I'm only using this on a subset of notebooks, but plan to expand over time. Also only using a subset of the linter checks, again with a view to revise and potentially expand over time.

Vale (nor any other spell checker / prose linter I could find) doesn't understand the notebook format, so `nb_vale.py` extracts the notebook markdown to a temp folder, then runs Vale on these files instead.

### To do:
- [x] Figure out how to make this work with CI

## How to read this PR

This looks like a lot of file changes but really `nb_vale.py` and `.github/workflows/test.yml` are the only logic changes, and follow the pattern of the rest of the content check scripts.
- The `./scripts/content_checks/style/` folder contains the Vale checks, these are pulled from other style guides and can be treated as an external library for review purposes.
- Changes to `./notebooks/` show examples of the kind of changes Vale catches. Only problem so far is Vale doesn't recognise a semicolon as a sentence break, so currently need to override.

Otherwise, try installing Vale, make some spelling mistakes, and watch it catch them.


vabarbosa (changes requested):
@frankharkins thank you for this. test is failing can you take a look? also, i am thinking maybe this would it make more sense as part of the pre-commit hook instead of CI test so errors are caught/resolved when attempting to merge rather than during CI test. thoughts?

and actually i'm thinking all notebooks checks could be part of pre-commit check rather than CI but we could discuss at a later time

frankharkins:
@vabarbosa thanks for taking a look! I think hooks would be nice, but anyone who modifies a notebook will need to have Vale installed to commit, is that ok? If so I can push that today.

vabarbosa:
> @vabarbosa thanks for taking a look! I think hooks would be nice, but anyone who modifies a notebook will need to have Vale installed to commit, is that ok? If so I can push that today.

true. i am ok with that and we could include it in the requirements-test.txt. however, you are more familiar with the notebook contributors so if you think it will slow down notebook contributions or is an unnecessary addition then i can be convinced to keep it in the CI tests as is

frankharkins:
@vabarbosa my hesitation is that it's not a Python package, but a standard executable. It's pretty easy to install, but still one more thing to remember / frustrate people. I propose waiting until most notebooks are covered in the lint checks.

vabarbosa (approved):
makes sense to me. thankyou!

PR #1247, opened 2022-06-29
Add QML course to lint checks

frankharkins:
To do:
- [ ] Write docstrings


vabarbosa (approved):
@frankharkins thank you!

PR #1261, opened 2022-07-01
Fix qiskit to 0.36

frankharkins:
## Changes

Qiskit 0.37 was released yesterday and has a few breaking changes. I will work on updating the notebooks, but for now fix the version to 0.36 so the checks can pass.


Tansito (approved):
Thanks for this @frankharkins !

PR #1269, opened 2022-07-02
Changing the figure order

ryota-katsube:
## Changes
I changed a sentence of code to match the order of the figures between  "images/qgan/real_rig.svg" and "images/qgan/fake_rig.svg" with the text sentences " Looking at the first circuit diagram shows us how real data samples are fed into the discriminator" and "In the second circuit, a generated [wave function](gloss:wave-function) aimed to ...".


Tansito:
Added @frankharkins to review the change! 🙏

frankharkins (approved):
Looks good, thanks!

PR #1272, opened 2022-07-03
Fix confusing math

derwind:
## Changes

Fixed confusing math

## Implementation details

Changed "... in the set $(i,j) = E$" to "... in the set $E$"
because $(i, j)$ is not $E$ itself but an element of it.

## Screenshots

qaoa


Tansito:
Added @frankharkins to review the change 🙏

frankharkins (approved):
Thank you!

Issue #1273, opened 2022-07-03
Bug: as of Qiskit 0.37.0 Ignis has been deprecated. VQE molecules notebook needs to be updated to account for that.


MarcoBarroca:
### Where does the issue happen?

https://qiskit.org/learn/course/quantum-algorithms-for-applications

### What is the content issue?

Qiskit Ignis has been deprecated. import list needs to be updated.

MarcoBarroca:
I'm not sure if there are other instances where inis is used in the notebook. Those would need to be updated as well.

PR #1274, opened 2022-07-03
Removed Ignis from import list in VQE molecules tutorial.

MarcoBarroca:
## Changes
Removed Ignis form the import list.

This should solve #1273. I had recently updated the notebook, no point in letting it become deprecated again.


Tansito:
Adding @frankharkins to the PR!

PR #1297, opened 2022-07-08
Fix code owners syntax

abdonrd:
## Changes

Now should work as expected

Reference: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file


Tansito (approved):
Thank you @abdonrd !

PR #1353, opened 2022-07-16
Added issue template for user research + future of feature to user st…

pandasa123:
## Changes

Extended our issue templates to include items from the Sprint Retro:

- Future of Feature (to `user_story.yml`)
- User Research Template (`user_research.yml`)

### DIFF
--- a/.github/ISSUE_TEMPLATE/user_research.yml
+++ b/.github/ISSUE_TEMPLATE/user_research.yml
@@ -1,0 +1,30 @@
+name: User Research
+description: Define new user research
+title: User Research for ...
+labels: ["user testing"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to create this research story!
+
+        In most cases, the Qiskit Textbook maintainers are the only ones creating and performing user research to test new ideas. If you believe you need to create a research task, please continue and provide as much information as you can.
+  - type: textarea
+    id: background
+    attributes:
+      label: Background
+      description: Set the scene and describe the motivation to conduct this user research
+  - type: textarea
+    id: audience
+    attributes:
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/user_research.yml
```diff
@@ -0,0 +1,30 @@
...
+    attributes:
+      label: Audience
+      description: High-level description of the intended audience 
+    validations:
+      required: true
+  - type: textarea
+    id: relevant
+    attributes:
+      label: Relevant existing research and issues 
+      description: Is
```
eddybrando, line 28: I don't understand this description. Was is unintended? If not, could you please help me understand what it means.

Thank you! 🙂

eddybrando (commented):
Thank you @pandasa123, this is super helpful!

Everything looks great, but I just have one small observation. Please help me clarify it and then we can merge this.

We could / should then also create / extend similar issues in our other repos.

.github/ISSUE_TEMPLATE/user_research.yml
pandasa123: Sure! There are two main use cases:

- Epics: sometimes we create user stories that are massive and should have multiple user stories attached, like Assignments and Problem Sets and Problem Set Discoverability
- Existing issues: sometimes we have old issues that are consolidated and formalised into a user story, like Equation Numbering

eddybrando (approved):
Thank you very much @pandasa123!

.github/ISSUE_TEMPLATE/user_research.yml
eddybrando: Resolved offline. `description: Is` can be removed.

PR #1356, opened 2022-07-19
Fix typos on unitaries

luukament:
## Changes
Remove vector arrows from several unitaries and improve display of subscripts on the same unitaries.


frankharkins (approved):
Looks good, thanks!

Issue #1344, opened 2022-07-14
Enable report on hardware executions


eddybrando:
As pointed out by @agebbie, we need to be able to report on hardware executions generated from the Textbook.

The strategy could be similar to what was deployed to distinguish local jobs from jobs run in using Quantum Labs: https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/api/session.py#L227

> Can we use the same header instrumentation used for jobs run in quantum labs so that when the binder is initialized we update the Qiskit provider header with [learn.qiskit.org](http://learn.qiskit.org/) for both the scratchpad and for all code cells running on hardware?

@Tansito pointed out:

> About the job it seems we are going to need to find a way to setup an environment variable in mybinder (probably using os.environ in the code-cell we will be able to do it, cc: @vabarbosa or @korgan00 for ideas here).

---

### Pull requests:

- [x] https://github.com/Qiskit/platypus-binder/pull/15
- [x] https://github.com/Qiskit/platypus/pull/1360

vabarbosa:
@eddybrando what additional info is needed here?

eddybrando:
> @eddybrando what additional info is needed here? it has the `needs information` label, just want to make sure i am not missing anything?

I think it's now clarified with [your comment](https://github.com/Qiskit/platypus/issues/1344#issuecomment-1189346386). 🙂 

We just needed to be sure about the way to proceed here.

vabarbosa:
after further discussion (in Slack), there are uses cases in the near future where option (1) would be preferred. 

with that said, i propose doing both! having both would ensure if there are any issues with (1) or if there are any pages where (1) is not in added then it would default to (2).

2- https://github.com/Qiskit/platypus-binder/pull/15
1- https://github.com/Qiskit/platypus/pull/1359

and we can easily close out either PR is anyone has any strong opinion against both

pandasa123:
This is working in staging! Excellent work everyone

PR #1360, opened 2022-07-20
execute init code in binder

vabarbosa:
## Changes

Updated binder initialization to set an env variable whose value will be appended to the header when making requests to hardware. the value of the env variable should be the location host & path from the current page

[Preview link](https://platypus-pr-1360.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/course/introduction/why-quantum-computing)

Related
- https://github.com/Qiskit/platypus/issues/1344
- https://github.com/Qiskit/platypus-binder/pull/15

## Implementation details

Added function to execute some initialization code (which set env variable) once binder is loaded/ready

## How to read this PR

review the function implementation & placement to make sure it is proper. then visit  multiple different pages and confirm the env variable added gets updated accordingly. this can be done by running the following in any code cell:

```python
import os
os.getenv('QE_CUSTOM_CLIENT_APP_HEADER')
```

## Screenshots

image


techtolentino (approved):
I've tested on several pages and this is working nicely for me. 

Not sure if anyone else wants to review but I'm good w/ it.

Nice work @vabarbosa 🚢

PR #942, opened 2022-04-26
Quantum 301 preview

frankharkins:
## Changes

This PR is to create a preview of [`jhwatrous:quantum-301`](https://github.com/jhwatrous/platypus/tree/quantum-301) to share with others. Unfortunately, #940 is skipping the preview, so I've pulled in the changes here.

[Link](https://platypus-pr-942.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/course/basics/)

PR #1367, opened 2022-07-26
Q301 Preview link for educators summit

frankharkins:
We need a temporary preview link for the 'Quantum 301' course to use at the educators summit. This is separate from #942 as the preview needs to be stable, but we don't want to block work on #942.


techtolentino:
@frankharkins - same here, do we need this branch now that we have https://learn.qiskit.org/course/basics/single-systems in production?

frankharkins:
No, thanks!

PR #1372, opened 2022-07-27
Add quantum user id to the qiskit user

### LINKED ISSUE #1339: Add in new IQX identifier column to Qiskit DB
### Feature description

Would like to request to add in a new column to the mongoDB Qiskit user table, which inserts the iqx user_id for users who have requested an API token, so that we can connect these users engagement on the textbook back to IQX, and understand full user journey.

### Additional context

_No response_

Tansito:
## Changes

Fix #1339 

## Implementation details

- Added to the `qiskit-user` model the property `quantumUserId` that will store the **id** for the quantum user.
- Created a repository entity in `libs` to connect with the quantum API and manage the different responses.
  - **Note**: I considered that this check could be a best effort and if something weird happens we just continue with the process and we don't return an error to the user.
  - **Note 2**: Added [got](https://github.com/sindresorhus/got) as NPM dependency to manage the HTTP call.
- In the service everytime the token is updated we check for the id of the user.

## How to read this PR

As we did with the old end-point we just need to pass the `apiToken` as we were doing and check if everything is working. It's good to mention that for staging we use the staging environment in quantum and the same way for production.

## Screenshots
Now the response will be:

```json
{
  "id": "62cfcf5d909f5",
  "apiToken": "182e0dd8897418430063a0665d5d8182e0dd88",
  "quantumUserId": "6270fe9721a29asdfghjklnm",
  "user": "6262c22345663dc8977ffe6a"
}
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "carbon-web-components": "^1.21.0",
         "class-validator": "^0.13.2",
         "codemirror": "^6.0.1",
-        "got": "^12.1.0",
+        "got": "^11.8.5",
         "lit-element": "^2.5.1",
         "lodash": "^4.17.21",
         "markdown-it": "^12.2.0",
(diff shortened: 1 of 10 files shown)


vabarbosa (approved):
thank you!

PR #1376, opened 2022-07-27
fix(ci): make "vale" available

### LINKED ISSUE #1373: Platypus not deploying due to Vale configuration
There is some issue with Vale's configuration that is making the deployment workflow fail in the test step.

https://github.com/Qiskit/platypus/runs/7425210455?check_suite_focus=true#step:10:111

eddybrando:
## Changes

This PR adds a step to the deployment workflow to install "Vale" before it get's used during testing, which caused the deployment to fail.

Closes #1373

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,6 +48,9 @@ jobs:
       - name: Install Python dependencies
         run: pip install -r converter/textbook-converter/requirements.txt -r converter/textbook-converter/requirements-test.txt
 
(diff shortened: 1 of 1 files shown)


frankharkins (approved):
LGTM, do we need to add this anywhere else?

eddybrando:
> LGTM, do we need to add this anywhere else?

Only the "release" and "test" workflow files have steps running the tests, so this should be it for now. 🙂

PR #1393, opened 2022-08-01
Correct indices in HHL page

### LINKED ISSUE #1377: Bug: HHL quantum solution doesn't match exact

https://qiskit.org/textbook/ch-applications/hhl_tutorial.html

On re-running the cells online, the HHL quantum solution in cell below doesn't match the exact solution.

Screenshot 2022-07-29 at 09 13 35

Changing indices 8 and 9 in below code snippet to 16 and 17 respectively seems to fix the issue.
```
from qiskit.quantum_info import Statevector

naive_sv = Statevector(naive_hhl_solution.state).data
tridi_sv = Statevector(tridi_solution.state).data

# Extract the right vector components. 1000 corresponds to the index 8 and 1001 corresponds to the index 9
naive_full_vector = np.array([naive_sv[8], naive_sv[9]])
tridi_full_vector = np.array([tridi_sv[8], tridi_sv[9]])

print('naive raw solution vector:', naive_full_vector)
print('tridi raw solution vector:', tridi_full_vector)
```

frankharkins:
## Changes

Corrected the indices as per suggestion in #1377. Looks like the circuit at some point aquired another qubit (maybe due to change in Qiskit), so we want the statevector indices corresponding to `10000` and `10001` instead of `1000` and `1001`.

Fixes #1377


eddybrando (approved):
Thank you @frankharkins!

PR #1399, opened 2022-08-03
Add hardware cell example to intro course

frankharkins:
## Changes

Adds a hardware cell to the intro course (resolves #1398)


pandasa123 (approved):
Amazing job @frankharkins!

notebooks/intro/why-quantum-computing.ipynb
```diff
@@ -242,8 +242,77 @@
...
+    "# 1. Create a simple quantum program called a 'quantum circuit'.\n",
+    "from qiskit import QuantumCircuit\n",
+    "qc = QuantumCircuit(3)\n",
+    "qc.h(0)\n",
+    "qc.cx(0, [1, 2])\n",
+    "qc.measure_all()\n",
+    "\n",
+    "# 2. Ask IBM Quantum for its least busy device that isn't a simulator.\n",
+    "from qiskit.providers.ibmq import IBMQ, least_busy\n",
+    "provider = IBMQ.load_account()\n",
```
pandasa123, line 266: I believe this line needs to be:

```python
provider = IBMQ.get_provider('ibm-q')
```

Since the following runs behind the scenes? 

```python
IBMQ.load_account()
```

pandasa123 (changes requested):

notebooks/intro/why-quantum-computing.ipynb
frankharkins: That will make the code run, but I think it would be better to fix the environment so `load_account` works correctly. Maybe that belongs in another issue though...

pandasa123: Yep that's by design since we didn't want users to think about that element in the Textbook

Pedagogically, we should add that as a comment?

```python
# If you're running this example locally, you need to load your account with your API token
# IBMQ.save_account(token="XYZ")
# IBMQ.load_account()
```

pandasa123 (approved):

PR #1401, opened 2022-08-04
add `is-api-token-needed` attribute

vabarbosa:
## Changes

pass the `is-api-token-needed` attribute from scratchpad to exercise action bar

## Implementation details

by default `is-api-token-needed` is true will disables run button. this PR update scratchpad component to set `is-api-token-needed` to false

this is an intermediate fix to re-enable the Run button. however, additional changes will be required to properly handle ardware execution in scratchpad

## How to read this PR

confirm scratchpad Run button is no longer disabled and code can be ran in the scratchpad

## Screenshots

image


techtolentino (approved):
Nicely done @vabarbosa 👏

PR #1418, opened 2022-08-08
Deleting cell that installs torchvision

alexao8:
## Changes

Deleted the cell which was installing torchvision in the "Hybrid quantum-classical Neural Networks with PyTorch and Qiskit" course. It's no longer needed because torchvision has been added to the platypus binder environment.


frankharkins (approved):
Awesome, thank you!

PR #1430, opened 2022-08-11
Minor fixes in introduction/entangled-states

### LINKED ISSUE #1371: [Content] Typos and minor suggestions

https://learn.qiskit.org/course/introduction/entangled-states

Incorrect (in my opinion) sentences:
- [x] "If we then apply only single qubit **states**, we will manipulate these individual single qubit states" shouldn't it be: "If we then apply only single qubit **gates**, we will manipulate these individual single qubit states"?
- [x] "However, there are ways that we can push the boundaries to do things that are not **possibly** classically" --> "However, there are ways that we can push the boundaries to do things that are not **possible** classically"
- [x] "Here we will look at **once** such example" --> "Here we will look at **one** such example"
- [x] "Similarly the **different** between" --> "Similarly the **difference** between"
- [x] "As such, while Alice just applying gates to one qubit, she was manipulating the larger set of four entangled pairs that it was a part of." - this doesn't seem correct to me. Alice was not manipulating the set of four entangled pairs of qubits. There was only one pair of qubits, and it could be in one of the 4 different states (while by state of a qubit pair we understand the states of both entangled qubits from that pair).

- [x] I have also one additional suggestion. In the screenshot I'm providing to this issue, there's introduced a subscript to the probability function name to denote this function provides the probability of measuring the qubit |a> as |0>. So it would be neat to denote the other probability function (that occurs twice on the right side of the equation) as p with subscripted |ba> to denote it as a different function that provides the probability of measuring the qubits |ba> as the corresponding argument value.

### Screenshots

subscripted_probability

frankharkins:
## Changes

Fixes #1371, see issue for full information

### DIFF
--- a/notebooks/intro/entangled-states.ipynb
+++ b/notebooks/intro/entangled-states.ipynb
@@ -187,7 +187,7 @@
     "\n",
     "Let's check this works: Suppose both qubits are measured, what is the probability that the qubit on the right comes out `0`? This will be the sum of the probabilities for the two outcomes which have a `0` on the right, namely `00` and `10`,\n",
     "\n",
-    "$$ \\cssId{_pa0}{p_{|a\\rangle}(|0\\rangle)} = p(|00\\rangle) + p(|10\\rangle). $$\n",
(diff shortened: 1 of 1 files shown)


vabarbosa (approved):
LGTM! thank you @frankharkins

PR #1451, opened 2022-08-18
Fix typo: add missing word

declanmillar:
## Changes

Adds a suspected missing word in the "What is quantum?" notebook:

"These cards were divided into grids, and *each* cell in the grid represents a bit."


frankharkins (approved):
Thank you!

PR #1452, opened 2022-08-18
fix typo: x -> cx

declanmillar:
## Changes

I think this is supposed to be a "CX" rather than "X" gate.
Also corrects a couple of typos in the text.


frankharkins (approved):
Thank you!

PR #1464, opened 2022-08-19
Fix a few typos in the text.

declanmillar:
## Changes

Just fixes a few typos in a text.


notebooks/intro/entangled-states.ipynb
```diff
@@ -1775,7 +1775,7 @@
...
+    "These correspond exactly to the observations made for the colored shapes in the classical scenario. If we apply the same logic as we did in that case, we can analyse the case where both Alice and Bob make an x measurement, and they both get the result `1`. Bob would then be able to apply observation 2 and infer that Alice would certainly have gotten a `0`, if she had instead done a z measurement. Alice can do the same, inferring that Bob would have gotten a `1` if he'd have made a z measurement. So together they conclude that, if they had both made z measurements, they would both have gotten the output `0`. Since observation 1 shows us that such a result would be impossible, Alice and Bob could predict that the `11` result will never occur.\n",
```
frankharkins, line 1778:
```suggestion
    "These correspond exactly to the observations made for the colored shapes in the classical scenario. If we apply the same logic as we did in that case, we can analyse the case where both Alice and Bob make an x measurement, and they both get the result `1`. Bob would then be able to apply observation 2 and infer that Alice would certainly have gotten a `0` if she had instead done a z measurement. Alice can do the same, inferring that Bob would have gotten a `1` if he'd have made a z measurement. So together they conclude that, if they had both made z measurements, they would both have gotten the output `0`. Since observation 1 shows us that such a result would be impossible, Alice and Bob could predict that the `11` result will never occur.\n",
```

frankharkins (dismissed):
Thank you!

frankharkins (approved):

PR #1474, opened 2022-08-23
VQC: Use the correct arXiv link for reference 3

rht:
Previously, the linked arXiv link has the paper title of "Quantum machine learning in feature Hilbert spaces."
It should have been "Supervised learning with quantum enhanced feature spaces."


frankharkins (approved):
Thank you!

PR #1484, opened 2022-08-26
Update Vale rules

frankharkins:
## Changes

- Added / tweaked a couple of Vale's rules
  - Change 'it is' rule to not match on end of sentence
  - Increase readability threshold; this should really apply to the whole notebook but at the moment we can only run it per cell, which leads to false positives
  - Add rule for pet peeve "reason why" vs just "reason"
- Added some more words to the dictionary

I made these changes while working on the quantum 301 branch, but think these belong in main.


eddybrando (approved):
Thank you very much @frankharkins! 🙂

PR #1503, opened 2022-09-01
Update Japanese toc.yaml 

kifumi:
## Changes
Updated [Japanese toc.yaml ](https://github.com/Qiskit/platypus/blob/main/translations/ja/toc.yaml)for adding translated and merged files


vabarbosa (approved):
thank you

PR #1504, opened 2022-09-01
Translation release guideline

kifumi:
## Changes
Add the guideline for the release of translation.

### DIFF
--- a/translations/release_guide.md
+++ b/translations/release_guide.md
@@ -36,7 +36,7 @@ Please see [toc.yaml for Japanese](https://github.com/Qiskit/platypus/blob/main/
 When the new translated files are merged, update toc.yaml file as follows.
 1. Copy the lines about new translated and merged files of [toc.yaml](https://github.com/Qiskit/platypus/blob/main/notebooks/toc.yaml) file under `./notebooks/`.
 2. Paste it in the correct position of your toc.yaml file under your language folder `./translations/xx/`.
-3. Translated the titles of the folders if you need.
+3. Translate the titles of the folders if you need.
 4. Remove the title lines of the file if it exists.
 5. Add `-` in front of `id`.
 6. Example: If the file `/ch-states/case-for-quantum` was newly merged, then `toc.yaml` should be changed like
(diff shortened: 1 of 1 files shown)


vabarbosa (approved):
thank you

ilyaspiridonov:
@kifumi @vabarbosa Quick follow-up on this: have you considered using Gitlocalize for the .yaml files? 
here's an experiment: https://gitlocalize.com/repo/7904/ja/notebooks/toc.yaml
and a sample output Gitlocalize generates: 
https://github.com/gitlocalize/platypus/pull/32/files

kifumi:
@ilyaspiridonov  toc.yaml works a kind of switcher which identifies whether the page has the language translation or not.  If the lines regarding the specific file are not exist on toc.yaml, the language switcher is disabled.  We need to delete the lines for which we didn't finish the translation.  So, I think it is better to modify it on github.  @vabarbosa how do you think about it?

ilyaspiridonov:
@kifumi Ah, got it, thanks. In theory, you would only translate the sections of the .yaml file that are needed (and leave the other ones untranslated) - which should still generate a yaml file, but only with the keys that were translated.

vabarbosa:
my apologies for the late reply. the `toc.yaml` is used to define 
- which notebooks to build & publish 
- what order they should be grouped & presented 
- the `title` to use in the navigation
- the `id` to use when generating a unique url to the content

given point (1) the `toc.yaml` has ended up being used like @kifumi suggested, as a way to identify which translations are available & should be published.

however, i was under the impression that GitLocalize was being used for all files to be translated (i.e, .ipynb, .yaml). in any case, if GitLocalize works like @ilyaspiridonov mentions then i dont see a reason not to use it.

> _only translate the sections of the .yaml file that are needed (and leave the other ones untranslated) - which should still generate a yaml file, but only with the keys that were translated_

ilyaspiridonov:
@vabarbosa thanks for the heads-up! Yeah, .yaml files are currently outside of Gitlocalize scope, and my initial though was to make everyone's life easier by processing those via Gitlocalize. But let me run some experiments first to see what the resulting toc.yaml would be

PR #1520, opened 2022-09-03
fix typo 1->0

hckim012:
## Changes

I think this is supposed to be a '0' instead of '1'.


vabarbosa:
@frankharkins agree?

frankharkins (approved):
Thank you

PR #1532, opened 2022-09-07
Fix #411

### LINKED ISSUE #411: [General UI] Strange vertical scrolling on content

1. Go to https://learn.qiskit.org/course/ch-gates/proving-universality#proving
2. Scroll down to the paragraph starting with "With a few Hadamards, we can do the same for" under Section 5
3. Scroll over the code example below the text and see an additional vertical scrolling bar.

When scrolling over the Qiskit code, there is an additional scrolling bar.

There is no need for the additional scrolling bar. Additionally that section should probably be contained in a code block.

### Screenshots (optional but very helpful)
Screen Shot 2021-11-04 at 10 19 04 AM

If applicable, please add screenshots or video captures to help explain the problem.

### Environment (optional but very helpful)
probably all

### Additional context (optional)

Add any other context that could help debug the problem.

ianhellstrom:
## Changes

Moved code in Markdown block to its own code cell with proper imports, so users can execute it.


frankharkins (approved):
Thank you!

PR #1545, opened 2022-09-11
Japanese fix for intro/what-is-quantum.ipynb

kifumi:
Fixed some issues in the transformation from GitLocalize.


0sophy1 (approved):
Thank you @kifumi ! Approve!

PR #1546, opened 2022-09-11
Japanese fix for intro/describing-quantum-computers.ipynb

kifumi:
Fixed issues from translation using GitLocalize.


0sophy1 (approved):
Thank you @kifumi !

PR #1548, opened 2022-09-11
Japanese fix for quantum-machine-learning/introduction.ipynb

kifumi:
Fixed issue from translation using GitLocalize.


ilyaspiridonov:
@kifumi I'm looking at today's fixes in translated notebooks. Can you please elaborate on this one - in this particular file was it something that prevented the build (critical)? Or is it just tidying up? I can only see indentation in this fix - is there something else that I'm missing?

kifumi:
@ilyaspiridonov Hi, this does not prevent the build, but it does not display correctly. Now it is upper figure below but correctly it should be bottom one.  We cannot correct it using GitLocalize.
image

0sophy1 (approved):
Thank you @kifumi . @ilyaspiridonov, our Japanese team undergoes several formula and indentation problems. I'll raise issues soon.

PR #1549, opened 2022-09-11
Japanese fix for quantum-machine-learning/encoding.ipynb

kifumi:
Fixed the issues from translation using GitLocalize.


0sophy1 (approved):
Thank you @kifumi ! Approved!

PR #1574, opened 2022-09-20
Add sanity checks to notebooks

frankharkins:
When creating notebooks, the writing should comment on the code outputs. When these notebooks are updated later on (maybe due to package updates or bugfixes), the code outputs can change, and the person updating the notebooks might not (and shouldn't really have to) know what a reasonable output looks like. E.g. for circuit results, is `516` close enough to `541`?. This PR adds the ability for writers to add 'sanity check' cells to notebooks that check the code outputs match what's written in the markdown cells.

## Changes

- Added logic in the converter to filter out code cells with `sanity-check` tags. (d033092b7f7edca7e28ec551cfdd94e6430efcc2)
- Added `_` and `Out` to Pylint builtins. (eef03b658d077e027b51e94b1ed77c89f00f2fe6)
- Added some sanity check cells to `intro/entangled-states`. I chose this notebook as I recognised an occurrance of code/prose mismatch (see screenshot). This notebook was also relatively simple acts as a good example for other contributors. (6d93e73e7148ae617b1ac71ffca37b87ef82956a)
  Screenshot 2022-09-20 at 17 01 34
- Also noticed that the CI runs all the notebooks instead of just the changed notebooks, hopefully have fixed that too. (20362b94d5d17a7def5eafc7e6dfaaafb995a329)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,7 +58,7 @@ jobs:
           npm run test:nb
           echo "::endgroup::"
           echo "::group::Run changed notebooks"
-          python3 ./scripts/content_checks/nb_autorun.py ${{ steps.changed-files.outputs.all_changed_files }} --fail-on-warning
+          python3 ./scripts/content_checks/nb_autorun.py ${{ steps.changed-notebooks.outputs.all_changed_files }} --fail-on-warning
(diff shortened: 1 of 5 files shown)


frankharkins:
Going to close while I

PR #1601, opened 2022-10-01
Fix incorrect symbols in formulas

derwind:
## Changes

Fixed incorrect symbols.

## Implementation details

Used `{latex}...` instead of `$...$` when using multiple underscores in formulae.

## Screenshots

Verification image after fix. Preview is done by `npm run buil`d & `npm start`.

- Before fix

before_fix

- After fix

after_fix


frankharkins (approved):
Thank you!

Issue #1602, opened 2022-10-02
Add Bengali and Chinese to Mathigon config


ilyaspiridonov:
### Feature description

Before translations can be made publishable, Bengali and Chinese need to be added (or we need to ensure these two are added) in Mathigon config on the server side under `node_modules/@mathigon/studio/server/data/locales.yaml`
**NB:** 
Bengali needs to be added with language code `bn`
Chinese is probably already there with the language code `cn` - needs to be replaced with `zh-CN` (to support the current path to Chinese translations in the repo) 

### Additional context

Mathigon config is not part of this repo (hence no pull request), however these two language codes need to be included (or someone needs to confirm they are already there on the server side – not the case by default), as otherwise the build will fail when these locales are added in [config.yaml](https://github.com/Qiskit/platypus/blob/main/config.yaml)

vabarbosa:
This repository is no longer maintained. It has been superseded by [IBM Quantum Learning](https://learning.quantum.ibm.com).

PR #1603, opened 2022-10-02
Make [KO],[ES], [PT] translations publishable

ilyaspiridonov:
Made Korean, Spanish and Portuguese translations publishable: 

- updated `config.yaml` to include the three languages in the build;
- added toc.yaml for each of the three languages. Section names are Google-translated.

**NB** Bengali and Chinese can not be added until [#1602](https://github.com/Qiskit/platypus/issues/1602) is resolved.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -5,7 +5,7 @@ domain: learn.qiskit.org
 # privacyURL: /policies
 # googleAnalytics: UA-XXX-XXX
 # banner: (c) Mathigon Ltd
-locales: [en, ja]
+locales: [en, ja, es, ko, pt]
 
 search:
   enabled: false
--- a/translations/es/toc.yaml
+++ b/translations/es/toc.yaml
@@ -1,0 +1,97 @@
+- title: Estados cuánticos y Qubits
+  url: /ch-states
+  sections:
+    - id: introduction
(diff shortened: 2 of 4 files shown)


clausia:
To publish each language, I thought we should follow this guide
https://github.com/Qiskit/platypus/blob/main/translations/release_guide.md

On the other hand, the translated toc.yaml file must be proofread before merging, isn't it?

ilyaspiridonov:
> 

Thank you @clausia,  this PR basically does what is described in the Release Guide. 
This is intended to lower the bar for the community to make new translated textbooks publishable (as right now the project has some notebooks available in 5 languages, but none of those have made it to the production release yet ;) 

Regarding proofreading - good point, I used Google Translate here as the initial translation, and the community can later improve if necessary. 
The translations are only the section titles (not textbook titles and TOC items, as those are parsed directly from the translated textbooks).

vabarbosa (approved):
@ilyaspiridonov thank you for getting this started. i switched the base branch to `feat/translations` (instead of `main`) that way we can set up/generate preview URL with the changes before pushing to main.

0sophy1:
@ilyaspiridonov Thank you so much..!

PR #1616, opened 2022-10-06
Fix latex blocks in lists

frankharkins:
Currently latex display equations do not work inside markdown lists. Currently we work around this using HTML tags within the markdown, but this PR makes a small modification to the converter (just preserving the indentation) to fix the problem.

## Changes

- Modified the converter to preserve indentation of latex blocks.
- Moved the few cases of latex inside lists to markdown instead of HTML.

## Additional details

Tests were failing in the density matrix page due to https://github.com/Qiskit/qiskit-terra/issues/8460, I've reset this notebook to skip the test


vabarbosa:
@frankharkins thanks. i've merged #1617 but this now has merge conflict which i didnt want to attempt to resolve (figured it best if you resolved)

vabarbosa (approved):
thank you!

PR #1617, opened 2022-10-06
fix Deutsch-Jozsa page

frankharkins:
## Changes

Fix some problems that prevented the Deutsch-Jozsa page from running correctly.

- Added the `qiskit-textbook` package as a requirement
- Removed an old widget that relied on an old version of Qiskit
- Added some sanity checks and refreshed the cell outputs


vabarbosa (approved):
LGTM

PR #1618, opened 2022-10-06
Fixed typo

derwind:
A parenthese is in the wrong position.

## Changes

- The closed parenthes is put outside of TeX notation.


vabarbosa (approved):
thank you

PR #1670, opened 2022-10-31
Update Qiskit version

frankharkins:
## Changes

Updates Qiskit to `0.39.0`, which includes a fix for some broken code on the
density matrix page.

Pylint raised a couple of import warnings, but the pages seemed to run fine in
the same environment so I've disabled them for now. If anyone knows why this
might be please let me know.


vabarbosa (approved):
thank you @frankharkins 
do we need to update the binder env as well?
https://github.com/Qiskit/platypus-binder/blob/main/environment.yml#L18

frankharkins:
Yes good catch! I've made a PR

PR #1674, opened 2022-11-01
Feature/curated syllabi

vabarbosa:
launch curated syllabi


vabarbosa (commented):
thank you @techtolentino @mariacloehb

PR #1676, opened 2022-11-02
Migrate notebooks to Qiskit/textbook

frankharkins:
Delete notebooks from this repo and include [Qiskit/textbook:stable/v2](https://github.com/Qiskit/textbook/tree/stable/v2) as a submodule.

This is in anticipation of moving to versioned releases; we need a single repo for the textbook content that is the source of truth for both the notebook content, and their metadata (e.g. urls, images, course descriptions).

> Note: I've left the summer school notebooks in Platypus (and not in [Qiskit/textbook](https://github.com/Qiskit/textbook/tree/stable/v2)).
>
> Although these are rendered by the "textbook platform", I don't think users would consider these pages as part of the textbook, especially as they appear below the tutorials in https://qiskit.org/learn/. I'm not tied to this and am open to other suggestions. I've left the problem sets notebooks out for these reasons too, but the logic is weaker for those.


frankharkins:
Closing in favour of a versioning solution

Issue #1675, opened 2022-11-02
Bug: Typo in section "Bell's Inequalities"


cjteclab:
### Where does the issue happen?

https://learn.qiskit.org/course/introduction/visualizing-entanglement

### What is the content issue?

In section "Bell's Inequalities/2.1 Bell test for classical variables" is a typo.
**has --> hash** 
Instead of:
"This is the probability that the bit value derived from an 'H' type hash on A is different to that from a 'V' type **has** on B."
It has to read:
"This is the probability that the bit value derived from an 'H' type hash on A is different to that from a 'V' type **hash** on B."

PR #1677, opened 2022-11-02
fix typo in section "Bell's Inequalities" (#1675)

cjteclab:
Replaced the word **has** with **hash** in section 2.1 Bell test for classical variables


frankharkins (approved):
Thanks!

PR #1686, opened 2022-11-09
Fix #1641

### LINKED ISSUE #1641: Bug: code snippet in why-quantum-computing only imports IBMQ after load account instructions

https://learn.qiskit.org/course/introduction/why-quantum-computing

I'm opening this issue on behalf of a user who reported this to me who would prefer to remain anonymous

The following code snippet provides instructions for `IBMQ.save_account()` but if a user uncomments this line it will break if they run it because` IBMQ` is only imported in the following line. It would be better for users if the import line was moved to before the comment about saving and loading the token

Screenshot 2022-10-18 at 11 47 07 AM

### Additional information

This is just one example that the user highlighted but they said they had many issues with code examples in this page (particularly because imports are missing or in the wrong place), so a general review of all the code snippets on the whole page is needed I think

frankharkins:
## Changes

Moved position of two commented-out lines, see #1641 for details.


vabarbosa (approved):
thanks

Issue #1684, opened 2022-11-08
Redirect syllabus duplication requests directly to `/edit/` screen


pandasa123:
### Feature description

Earlier we decided to ship the MVP to redirect a user to their `/account/classroom` page. After inspecting the data, we should instead route them to the `/edit/<SYLLABUS-CODE>` page directly

### Additional context

[Original PR](https://github.com/Qiskit/platypus/pull/1674/) by @techtolentino

PR #1688, opened 2022-11-09
feature(curated-syllabi): update redirect

techtolentino:
## Changes

Fixes https://github.com/Qiskit/platypus/issues/1684

## Implementation details

- when a user copies a syllabus, they are redirected to the `syllabus/edit/<SYLLABUS-CODE>` form
  - updated the form so the Syllabus name persists with the "Copy of [Original syllabus name]" convention
  - keeps the first field (Syllabus name) in focus, so the user has an indication of what they just duplicated
  

## Screenshots

https://user-images.githubusercontent.com/6276074/200942415-38c491fe-1bfb-4cc2-891f-d79edb0bb851.mov

### DIFF
--- a/frontend/vue/components/Syllabus/SyllabusForm.vue
+++ b/frontend/vue/components/Syllabus/SyllabusForm.vue
@@ -109,11 +109,6 @@ export default defineComponent({
     if (activeSyllabus) {
       this.syllabus = activeSyllabus
       this.editMode = true
-
-      if (activeSyllabus.name.startsWith('Copy of')) {
-        this.syllabus.name = ''
-        this.syllabus.instructor = ''
-      }
     }
     this.syllabusInfoChanged(this.syllabus)
   },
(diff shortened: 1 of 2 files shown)


vabarbosa (commented):
@techtolentino thanks! this looks good. just one comment... should we at least clear/delete the instructor name when going directly to edit page?

techtolentino:
@vabarbosa 

That's a great point... 

I thought about that, but it's too early to tell if we should clear it or not. 

I pseudo-tested this by clearing it locally, then tried to put myself in the experience of a new user, and I could totally see a user missing that field, only to realize later when they want to hit "Publish", that they need to go back up and fill in their Instructor name.

My guess is that if a user is prompted to rewrite the title field, they'll notice _hey, I'm not Peter Shor... and since I'm editing the Syllabus title, let me go ahead and update the rest of the fields_ 

(✋🏼 this is definitely an assumption)

Still, I think it's too early to know either way.

I'd vote ship what we have here, and if we need to clear it, we can make a new PR?

vabarbosa (approved):
thank you @techtolentino 
waiting to see if we need to clear instructor works for me.

PR #1719, opened 2022-11-27
final code block missing bokeh package

anujang:
for final code block regarding classical vs quantum code block, bokeh package is not imported. Bokeh can be installed in the textbook page by pip installing manually. which means on lines 186 and 187, plot function has be altered slightly to work with newer bokeh package


frankharkins (approved):
Thank you! I've re-run the cell to render the widget, and added the bokeh package to the dependencies (and to the [binder environment](https://github.com/Qiskit/platypus-binder/pull/27)). Ideally the code that generates the widget is hidden from the user, but that's out of scope of this PR.

PR #1720, opened 2022-11-27
Multi-qubit problem sets

amirhassank:
## Changes

Added multi-qubit problem sets


notebooks/problem-sets/multi-qubit_states.ipynb
```diff
@@ -0,0 +1,380 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "goals": [
+     {
+      "id": "multi-qubit-states-2"
+     }
+    ],
+    "grader_answer": "qc",
+    "grader_id": "textbook_practice_exam/multi-qubit_states/problem2"
```
vabarbosa, line 59: the `grader_id` should be in the form of `<event or exam>/<problem id>` so therefore this should be 
`textbook_practice_exam/problem2` or better yet to avoid possible conflict with other problems in `textbook_practice_exam` maybe `textbook_practice_exam/multi-qubit-problem2` would be a better `grader_id` value

notebooks/problem-sets/multi-qubit_states.ipynb
```diff
@@ -0,0 +1,380 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "goals": [
+     {
+      "id": "multi-qubit-states-7"
+     }
+    ],
+    "grader_answer": "qc",
+    "grader_id": "textbook_practice_exam/multi-qubit_states/problem7"
```
vabarbosa, line 237: the `grader_id` should be in the form of `<event or exam>/<problem id>` so therefore this should be
`textbook_practice_exam/problem7` or better yet to avoid possible conflict with other problems in `textbook_practice_exam` maybe `textbook_practice_exam/multi-qubit-problem7` would be a better `grader_id` value

vabarbosa (changes requested):
@amirhassank thank you, just a couple quick comment regarding the `grader_id` value

vabarbosa:
@amirhassank are you able to sign the CLA? also are you able to address the comments above? thanks. let me know if you have any questions.

PR #1722, opened 2022-11-28
feat: update issue templates and workflows

eddybrando:
## Changes

With this change, we're removing the auto assigning for new issues. As previously discussed, assigning somebody to an issue should be more intentional.

The label "needs grooming" and the workflow to move all new issues to the organisation board are also removed. Now, new issues will remain initially in their repositories until they are manually checked and added to the organisation board, which will put them in our organisation backlog. From there, they can be easily added to a future sprint for better planing.


techtolentino (approved):
Similar update to the one in the qiskit org repo

LGTM

vabarbosa (approved):
thank you!

PR #1697, opened 2022-11-17
Publish multiple systems page

frankharkins:
This is ready for review! This PR:
- publishes the new multiple systems page,
- splits the _circuits, protocols, and games_ page into two pages, and
- adds new words to the spellchecker.

PR #1723, opened 2022-11-28
Handle markdown blockquotes

frankharkins:
At the moment, we have no way of handling markdown blockquotes (i.e. lines starting with `>`). This PR adds that functionality.

## Changes
 
- Added logic in the converter to recognise lines starting with `>` as block quotes, and to wrap them in `<blockquote>` tags.
- Added scss to indent these blocks slightly, and add a left margin (similar to GitHub & Jupyter notebook)

## Screenshots

The code below is from the multiple systems page of the basics course (see #1697):

```
For example, suppose that $\mathsf{X}$ and $\mathsf{Y}$ are bits, and consider an operation with the following description.

> If $\mathsf{X} = 1$, then perform a NOT operation on $\mathsf{Y}$, otherwise do nothing.

This is a deterministic operation known as a *controlled-NOT* operation, where $\mathsf{X}$ is the *control* bit that determines whether or not a NOT operation should be applied to the *target* bit $\mathsf{Y}$.
```

and renders like this:

Screenshot 2022-11-28 at 14 00 26

previously, the content in the block quote did not appear at all.

## Other examples

This is how GitHub handles block quotes

> This is a block quote

Similarly in notebooks:

Screenshot 2022-11-28 at 15 07 28

### DIFF
--- a/converter/textbook-converter/textbook_converter/TextbookExporter.py
+++ b/converter/textbook-converter/textbook_converter/TextbookExporter.py
@@ -245,6 +245,7 @@ def handle_markdown_cell(cell, resources, cell_number, is_problem_set=False):
     markdown_lines = []
     lines = cell.source.splitlines()
     in_latex = False
+    in_blockquote = False
     in_block = False
     in_code = False
     headings = []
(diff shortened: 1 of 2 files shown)


JRussellHuffman (approved):

JRussellHuffman:
LGTM. thanks for this.

vabarbosa (approved):
thank you @frankharkins 
so we dont lose this for saiba perhaps you can create an issue in [that repo](https://github.com/Qiskit/saiba) so we make similar update (or if you prefer you can try to submit the PR itself there as well)

PR #1738, opened 2022-12-03
update Japanese toc.yaml

kifumi:
## Changes
Updated `./translations/ja/strings.yaml` based on the resent translation.


0sophy1 (approved):
Thank you so much @kifumi !

PR #1744, opened 2022-12-05
Add illustrations to multiple systems page

frankharkins:
## Changes

- Added new illustrations to the multiple systems page
- Re-organised existing images


vabarbosa (approved):
thank you!

PR #1751, opened 2022-12-07
Replace images in multiple systems page with higher res versions

frankharkins:
The images uploaded in #1744 were not the highest resolution versions, this PR replaces them.


vabarbosa (approved):
thank you @frankharkins

PR #1759, opened 2022-12-09
Use 'uses-hardware' tag to check if cell requires api token

frankharkins:
At the moment, we disable any code cells that contain '`IBMQ.`' as they probably connect to IBM Quantum, and can't run without an API token. This does not stop users from running cells further down the notebook that rely on results from IBM Quantum. Fortunately, most notebooks contain '`uses-hardware`' tags for cells that rely on results from IBM Quantum, this PR uses those tags to block more cells.

## Changes

- Modify converter to pass the `uses-hardware` tag through
- Modify the component to use this property to determine if a cell needs an API token
- Modify the messages to closer reflect the reason we've disabled the cell

## Screenshots
From [Deutsch-Jozsa algorithm page](https://learn.qiskit.org/course/ch-algorithms/deutsch-jozsa-algorithm#device) (left before, right after).
<div style="display: grid; grid-template-columns: 1fr 1fr;">
  
  
</div>

### DIFF
--- a/converter/textbook-converter/textbook_converter/TextbookExporter.py
+++ b/converter/textbook-converter/textbook_converter/TextbookExporter.py
@@ -365,27 +365,31 @@ def handle_code_cell_output(cell_output):
     return None
 
 
-def handle_grader_metadata(cell_metada):
-    """Parse grader metadata and return code exercise widget syntax
+def handle_code_cell_metadata(cell_metada):
+    """Parse code cell metadata (including grader information) and return code
+    exercise widget syntax
     """
-    grader_attr = None
+    attr = ''
 
     if "grader_import" in cell_metada and "grader_function" in cell_metada:
         grader_import = cell_metada["grader_import"]
         grader_function = cell_metada["grader_function"]
-        grader_attr = f'grader-import="{grader_import}" grader-function="{grader_function}"'
+        attr = f'grader-import="{grader_import}" grader-function="{grader_function}"'
     elif "grader_id" in cell_metada and "grader_answer" in cell_metada:
         grader_id = cell_metada["grader_id"]
         grader_answer = cell_metada["grader_answer"]
-        grader_attr = f'grader-id="{grader_id}" grader-answer="{grader_answer}"'
+        attr = f'grader-id="{grader_id}" grader-answer="{grader_answer}"'
 
-    if grader_attr:
+    if attr:
         goal = cell_metada["goals"] if "goals" in cell_metada else None
 
         if goal is not None:
-            grader_attr = f"{grader_attr} goal=\"{goal[0].id}\""
+            attr = f"{attr} goal=\"{goal[0].id}\""
 
-    return f"q-code-exercise({grader_attr or ''})"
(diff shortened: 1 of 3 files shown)


converter/textbook-converter/textbook_converter/TextbookExporter.py
```diff
@@ -365,27 +365,31 @@ def handle_code_cell_output(cell_output):
...
-        grader_attr = f'grader-id="{grader_id}" grader-answer="{grader_answer}"'
+        attr = f'grader-id="{grader_id}" grader-answer="{grader_answer}"'
 
-    if grader_attr:
+    if attr:
         goal = cell_metada["goals"] if "goals" in cell_metada else None
 
         if goal is not None:
-            grader_attr = f"{grader_attr} goal=\"{goal[0].id}\""
+            attr = f"{attr} goal=\"{goal[0].id}\""
```
frankharkins, line 387:
```suggestion
            attr += f" goal=\"{goal[0].id}\""
```

frontend/vue/components/UtilityPanel/Scratchpad.vue
```diff
@@ -29,11 +29,11 @@
       <div v-if="isApiTokenNeeded" class="scratchpad__api-token-info">
         <WarningIcon />
         <div>
-          This code is executed on real hardware using an IBM Quantum provider. Setup your API-token in
+          This code looks like it connects to IBM Quantum. Add your API-token to
           <BasicLink url="/account/admin">
             your account
           </BasicLink>
-          to execute this code cell.
+          to run this cell.
```
vabarbosa, line 36: i would just run this by design (@JRussellHuffman , @gracelindsell) and confirm they are OK with the updated text.

vabarbosa (approved):
thank you @frankharkins 
this is great. i have approved, but just have one comment before you merge

frontend/vue/components/UtilityPanel/Scratchpad.vue
frankharkins: Sounds good, thanks. That message is just for the scratchpad (which only checks for `IBMQ.`, the message for the inline cells is a bit different:

> This code connects to, or relies on results from IBM Quantum. Add your API-token to your account to run this cell.

JRussellHuffman: What is the updated text you are considering? Is it the text in the screenshots above, or @frankharkins's quoted text directly above, or both?

frankharkins: @JRussellHuffman 
For scratchpad:
> This code looks like it connects to IBM Quantum. Add your API-token to your account to run this cell.

For inline cells:
> This code connects to, or relies on results from IBM Quantum. Add your API-token to your account to run this cell.

The screenshot only shows the text for the inline cells.

vabarbosa: @JRussellHuffman @gracelindsell you good with the above text updates.

JRussellHuffman: yep. I like it.

PR #1766, opened 2022-12-14
[minor] fix some textbook errors and update README

patriciali:
1. Typo under "Single systems" section 2.2.3

Before:
Screenshot_12_13_22__7_53_PM

After:
Screenshot_12_13_22__7_54_PM

2. Error under "Multiple systems" section 1.1.4: the matrix representation of the operation has the correct shape, but it has dimension 9. It should have dimension 8, because 2^3 = 8.

Before: 
Screenshot_12_14_22__10_12_PM

After:
Screenshot_12_14_22__10_13_PM

3. Error under "Multiple systems" section 1.1.4: it says b, d in \Gamma, but it should be **c**, d in \Gamma.

4. I also updated the dev URL in the README to http instead of https because there is no SSL cert for dev.


frankharkins (approved):
Awesome, thank you!

PR #1788, opened 2023-01-04
Add rules to catch name typos

frankharkins:
## Changes

Adds rules to catch some bad typos I've seen elsewhere.


vabarbosa (approved):
thank you! 👍🏽

PR #1799, opened 2023-01-08
Update required Node version in README

### LINKED ISSUE #1798: Bug: Building the textbook required Node v16 and WSL for a Windows user

1. Download Node v18 (the current LTS) for Windows
2. Attempt to follow the steps in the README to build platypus

The python requirements will fail to download on Windows. If one switches to WSL then the python requirements will download, but numerous issues are given for the `npm install` command. One needs to use WSL and use `nvm` to choose Node v16 for everything to work. 

While the node errors make it clear that Platypus will not work with Node v18, the README says that Node v14 or later would work, which it doesn't. It also says nothing about Windows not being usable.

### Screenshots

image

### Device

Laptop

### OS

Windows 11

### Browser

_No response_

### Additional information

I imagine a temporary fix would be to update the README while platypus gets updated to work with later versions of node.

micpap25:
Fixes #1798

Updates the README to specify the correct required version of Node to build Platypus.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ The textbook is intended for use as a university quantum algorithms course suppl
     $ pip install -r requirements.txt -r requirements-test.txt
     ```
 
-1. Node.js (14.0 or later) (you could install node.js from [here](https://nodejs.org/en/download/)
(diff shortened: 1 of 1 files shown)


vabarbosa (approved):
thank you @micpap25

PR #1802, opened 2023-01-09
Fix broken underscores by forcing MathJax to parse

### LINKED ISSUE #1801: Bug:  Malformed formula in Simon's Algorithm page due to underscore misinterpretation by Markdown

https://learn.qiskit.org/course/ch-algorithms/simons-algorithm

A formula in section 2 is malformed and does not render correctly on the website despite rendering correctly in GitHub.

### Screenshots

What it looks like
image

What it should look like
image

### Additional information

This is caused by Markdown reading the underscores in the TeX formula section as emphasis rather than subscript (their intention).

I am not sure why it happens for this specific pair but I will work RN on fixing it.

micpap25:
Fixes #1801

Uses code brackets to prevent Markdown from turning underscores into emphasis, allowing MathJax to properly render this equation. This should be done for all inline equations with multiple underscores; I couldn't find other instances of this occurring after a brief scan so let me know if there is a known way to solve this issue.


frankharkins (approved):
Thank you! Seems to work correctly.

Before:
Screenshot 2023-01-09 at 19 23 00

After:
Screenshot 2023-01-09 at 19 23 36

vabarbosa (approved):
thank you @micpap25

PR #1804, opened 2023-01-10
Fix Japanese translation

kifumi:
Updated the wrong appearance of the equations on Japanese translation of "ch-states/representing-qubit-states.ipynb."


vabarbosa (approved):
thank you

PR #1805, opened 2023-01-10
Fix Japanese translation: ch-states/single-qubit-gates

kifumi:
Updated the wrong appearance of the equations on Japanese translation of "ch-states/single-qubit-gates.ipynb."


vabarbosa (approved):
thank you

PR #1806, opened 2023-01-10
Fix Japanese translation: ch-gates/multiple-qubits-and-entangled-states

kifumi:
Updated the wrong appearance of the equations on Japanese translation of "ch-gates/multiple-qubits-and-entangled-states.ipynb."


0sophy1 (approved):
Thank you @kifumi for your contributions!

PR #1808, opened 2023-01-10
Fix Japanese translation: ch-algorithms/quantum-teleportation

kifumi:
Updated the wrong appearance of the equations on Japanese translation of "ch-algorithms/quantum-teleportation.ipynb."


0sophy1 (approved):
Thank you so much @kifumi for your contributions!

PR #1820, opened 2023-01-14
enhance import code block and refactor with function for VQE

### LINKED ISSUE #1224: Improve code in VQE page

https://learn.qiskit.org/course/ch-applications/simulating-molecules-using-vqe

Thanks to #1188 the code in the VQE page now works correctly, but there are still places it can be improved.

- [ ] Neaten imports: The imports cell is huge, we might be able to import higher-level modules instead of importing each class individually, i.e. `import qiskit`, then use `qiskit.algorithms.optimizers.SLSQP(...)` in the main code. This will depend on how frequently we use each class.
- [ ] The functions have a lot of local variables, maybe we can reduce this or split into smaller functions.
- [ ] Need to add docstrings to functions.

SamYuan1990:
Signed-off-by: Sam Yuan <yy19902439@126.com>

## Changes

enhance import code block and refactor with function for VQE
fix #1224

## Implementation details

Neaten imports: The imports cell is huge, we might be able to import higher-level modules instead of importing each class individually, i.e. import qiskit, then use qiskit.algorithms.optimizers.SLSQP(...) in the main code. This applied for import less than 3 times.
refactor function get_qubit_op to make it reusable.


frankharkins:
@SamYuan1990 thanks for this! I've tried running but am getting an error:

```
AttributeError: module 'qiskit_nature' has no attribute 'transformers'
```

This is with both `qiskit_nature==0.4.0` and `0.5.1`. Which version are you using?

SamYuan1990:
https://github.com/Qiskit/platypus/blob/main/converter/textbook-converter/requirements-test.txt#L7 @frankharkins

SamYuan1990:
convert to draft to avoid notifications msg.

SamYuan1990:
```
yuanyi@yuanyideMacBook-Pro platypus % /opt/local/bin/python3.9 ./scripts/content_checks/nb_autorun.py ch-applications/vqe-molecules.ipynb

- [22:55] notebooks/ch-applications/vqe-molecules.ipynb 

✔
Finished in 321.74 seconds
```
it seems python 3.8 and python 3.9 returns different result in lint checking, hence turn this PR to draft and move testing to pipeline running with my local fork for development.

SamYuan1990:
@frankharkins , could you please help trigger CI and let's see if it works?
so far it works on my own fork's pipeline.
https://github.com/SamYuan1990/platypus/actions/runs/3969633978/jobs/6804385425

frankharkins:
Hi, thanks for taking a look, CI is running now. We currently support Python 3.8 and Qiskit Nature 4.0.0, although I'd be happy to update Qiskit Nature to latest if needed. I intended to investigate myself before I replied (as I couldn't contribute anything useful), but didn't get the chance.

woodsp-ibm:
Some comments. Main one is that, to me, this really could do with updating to latest versions. 

For example Aer should now be imported from qiskit_aer. The namespace integration, which allowed it to be imported from qikist, is deprecated and will be removed.

For VQE this is using the now legacy version which is pending deprecation. Since Qiskit 0.22.0 there are new/refactored algorithms, including VQE, that are based on the new primitives, Estimator and Sampler.

The Qiskit Nature logic here is using now deprecated code too. There is [a migration guide](https://qiskit.org/documentation/nature/migration/index.html) to help alter code to the latest 0.5..

I don't think the code will stop working for while yet, but maybe an issue should be raised to looking into this.

One bit of text that ought to be fixed right away though is this:

>This section illustrates an implementation of VQE using the programmatic approach. Qiskit also enables a declarative implementation, however, it reveals less information about the underlying algorithm. This code, specifically the preparation of qubit operators, is based on the code found in the Qiskit Tutorials repository (and as of July 2019, may be found at: https://github.com/Qiskit/qiskit-tutorials ).

While Qiskit Aqua (code has long been migrated elsewhere and is no longer supported) did have a declarative mechanism that was anyway removed some years ago and left only Python programming APIs before it was discontinued. As to the references  to prep of operators maybe there is some tutorial in the present docs that serves as a better link rather than generically to qiskit-tutorials as they were in 2019.

SamYuan1990:
> Some comments. Main one is that, to me, this really could do with updating to latest versions.
> 
> For example Aer should now be imported from qiskit_aer. The namespace integration, which allowed it to be imported from qiskit, is deprecated and will be removed.
> 
> For VQE this is using the now legacy version which is pending deprecation. Since Qiskit 0.22.0 there are new/refactored algorithms, including VQE, that are based on the new primitives, Estimator and Sampler.
> 
> The Qiskit Nature logic here is using now deprecated code too. There is [a migration guide](https://qiskit.org/documentation/nature/migration/index.html) to help alter code to the latest 0.5..
> 
> I don't think the code will stop working for while yet, but maybe an issue should be raised to looking into this.
> 
> One bit of text that ought to be fixed right away though is this:
> 
> > This section illustrates an implementation of VQE using the programmatic approach. Qiskit also enables a declarative implementation, however, it reveals less information about the underlying algorithm. This code, specifically the preparation of qubit operators, is based on the code found in the Qiskit Tutorials repository (and as of July 2019, may be found at: https://github.com/Qiskit/qiskit-tutorials ).
> 
> While Qiskit Aqua (code has long been migrated elsewhere and is no longer supported) did have a declarative mechanism that was anyway removed some years ago and left only Python programming APIs before it was discontinued. As to the references to prep of operators maybe there is some tutorial in the present docs that serves as a better link rather than generically to qiskit-tutorials as they were in 2019.

I just have 1 question for this. well, we can update dependencies for VQE, but if we do upgrade, is there any impacts for rest of notes?

woodsp-ibm:
>I just have 1 question for this. well, we can update dependencies for VQE, but if we do upgrade, is there any impacts for rest of notes?

The new Estimator based VQE is only supported by the new Nature 0.5.0 so that respective code would need to be updated in this notebook too. The driver/problem aspect of Nature was refactored for 0.5.0 to facilitate better integration with classical chemistry codes and allow such code to drive the problem from that level. It was a significant refactor at that level and the new  Nature code now sits in qiskit_nature.second_q. The migration guides and tutorials show how to migrate and have examples for ground state solving. I also see this notebook is using opflow from Terra, specifically the two qubit reduction logic. Since this is done by most users via the QubitConvertor/ParityMapper it can be specified there rather than doing this explicitly.  I say that since opflow in Terra will be deprecated and removed in the near future (had been intended to be done this imminent release of Terra but was held back until the next release). Two qubit reduction code, since it was Nature specific is being moved to Nature and the Parity Mapper - in fact this rework has already been done for the next Nature release knowing opflow will be deprecated. I will also note that Estimator's have been created to abstract away  computing the expectation value of an operator. Yes you can still get ideal outcome from Terra and  Aer versions that are equivalent to using state_vector simulator. The execution of circuits for an Estimator on a qasm like simulator or real-device is likely to done as stated, i.e Paulis are measured - in practice with Nature operators can have hundreds of Paulis and executing a circuit for each is expensive. So by default now, and in fact for a long time, Paulis are grouped into commuting sets so a single measurement can be applied to a whole group and far less run. With the runtime supporting additional capability of course more circuits may be run, e.g. for noise mitigation. seems this notebook covers a bunch of different aspects at different details.

SamYuan1990:
> > I just have 1 question for this. well, we can update dependencies for VQE, but if we do upgrade, is there any impacts for rest of notes?
> 
> The new Estimator based VQE is only supported by the new Nature 0.5.0 so that respective code would need to be updated in this notebook too. The driver/problem aspect of Nature was refactored for 0.5.0 to facilitate better integration with classical chemistry codes and allow such code to drive the problem from that level. It was a significant refactor at that level and the new Nature code now sits in qiskit_nature.second_q. The migration guides and tutorials show how to migrate and have examples for ground state solving. I also see this notebook is using opflow from Terra, specifically the two qubit reduction logic. Since this is done by most users via the QubitConvertor/ParityMapper it can be specified there rather than doing this explicitly. I say that since opflow in Terra will be deprecated and removed in the near future (had been intended to be done this imminent release of Terra but was held back until the next release). Two qubit reduction code, since it was Nature specific is being moved to Nature and the Parity Mapper - in fact this rework has already been done for the next Nature release knowing opflow will be deprecated. I will also note that Estimator's have been created to abstract away computing the expectation value of an operator. Yes you can still get ideal outcome from Terra and Aer versions that are equivalent to using state_vector simulator. The execution of circuits for an Estimator on a qasm like simulator or real-device is likely to done as stated, i.e Paulis are measured - in practice with Nature operators can have hundreds of Paulis and executing a circuit for each is expensive. So by default now, and in fact for a long time, Paulis are grouped into commuting sets so a single measurement can be applied to a whole group and far less run. With the runtime supporting additional capability of course more circuits may be run, e.g. for noise mitigation. seems this notebook covers a bunch of different aspects at different details.

ok, may I have a list of libraries which I'd better to do upgrade.
Recently I am on my Chinese New Year holiday, I update those libraries when I am available.

woodsp-ibm:
> ok, may I have a list of libraries which I'd better to do upgrade.

For me this would be to use the new logic in

- Nature 0.5.0 (most code now is in qiskit_nature.second_q so pretty much all the imports for it would have that)
- Terra 0.22 (or above - 0.23 will be released very shortly) use the Estimator based VQE. You can see the new VQE being used here in the[ algorithm tutorials](https://qiskit.org/documentation/tutorials.html#algorithms) for how its used. (Its much like the former VQE but it takes an Estimator - Terra has one in qiskit.primitives, and Aer has one too. I believe the tutorials include use of both - look at the first 3 of them). Runtime has Estimators too, with noise mitigation etc. But I imagine running that might be outside the scope for the textbook. For Aer make sure you use 0.11.2  (or above) if you run with an Estimator from Aer as it fixed some issues when use in conjunction with the algorithms.
- Lastly, though this is not really new library related as such, I would avoid using any `qiskitopflow`, like the tutorial does for 2 qubit reduction, since as I mentioned above this code will be deprecated and removed in the near future.

frankharkins (dismissed):
@SamYuan1990 thanks for this. Since this PR closes its linked issue (https://github.com/Qiskit/platypus/issues/1224), I'm going to merge it.

I've created a separate issue (#1840) for the updates suggested by @woodsp-ibm.

frankharkins (approved):

PR #1822, opened 2023-01-15
Fix Japanese translation: ch-applications/simulating-molecules-using-vqe

kifumi:
Updated the wrong appearance of the equations on Japanese translation of "ch-applications/simulating-molecules-using-vqe.ipynb."


0sophy1 (approved):
Thank you @kifumi !

PR #1823, opened 2023-01-15
Fix Japanese translation: quantum-machine-learning/kernel

kifumi:
Updated the wrong appearance of the equations on Japanese translation of "quantum-machine-learning/kernel.ipynb."


0sophy1 (approved):
Thank you @kifumi for your efforts! :)

PR #1832, opened 2023-01-18
Handle display equations in blockquotes

frankharkins:
## Changes

Tweaks some converter logic to handle display equations inside blockquotes. This problem appears in the new basics page.

## Implementation details

Markdown blockquotes are indented with a chevron (`>`). Since the converter detects display equations as lines starting with `$$`, display equations inside blockquotes were not detected. The converter also ends a blockquote when lines are no longer indented with `>`.

This PR strips left chevrons in the same way as whitespace so display equations are detected, and will not close blockquotes until the display equation has ended.

## Screenshots

Before:

Screenshot 2023-01-18 at 13 57 54

After:

Screenshot 2023-01-18 at 13 58 22

The markdown:

```markdown

The *no-cloning theorem* shows it is impossible to create a perfect copy of an unknown quantum state.

> **Theorem (No-cloning theorem)**:
>
> Let $\mathsf{X}$ and $\mathsf{Y}$ be systems sharing the same classical state set $\Sigma$ having at least two elements. There does not exist a quantum state $\vert \phi\rangle$ of $\mathsf{Y}$ and a unitary operation $U$ on the pair $(\mathsf{X},\mathsf{Y})$ such that
>
> $$
    U \bigl( \vert \psi \rangle \otimes \vert\phi\rangle\bigr)
    = \vert \psi \rangle \otimes \vert\psi\rangle
    \tag{6}
  $$
>
> for every state $\vert \psi \rangle$ of $\mathsf{X}$.
 
```

### DIFF
--- a/converter/textbook-converter/textbook_converter/TextbookExporter.py
+++ b/converter/textbook-converter/textbook_converter/TextbookExporter.py
@@ -262,8 +262,9 @@ def handle_markdown_cell(cell, resources, cell_number, is_problem_set=False):
                 markdown_lines.append("\n")
                 in_latex = True
             continue
-        if line.lstrip().startswith("$$"):
+        if line.lstrip(' >').startswith("$$"):
             indent, l = line.split("$$", 1)
+            indent = indent.replace('>', '')
             assert not indent or indent.isspace()
             markdown_lines.append(f"{indent}```latex\n")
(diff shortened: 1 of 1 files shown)


vabarbosa (approved):
thank you!

PR #1841, opened 2023-01-23
Correct branding

frankharkins:
## Changes

- IBM Q -> IBM Quantum
- qiskit -> Qiskit
- Added a check for both these cases


vabarbosa (approved):
thank you

PR #1847, opened 2023-01-25
Use runtime in 'Why quantum computing?'

frankharkins:
## Changes

Adapts the code example to use runtime primitives.


vabarbosa (approved):
thank you

PR #1864, opened 2023-01-29
Fix Japanese translation: ch-states/atoms-computation.ipynb

kifumi:
Updated the wrong appearance of the equations on Japanese translation of "ch-states/atoms-computation.ipynb".


veenaiyuri (approved):
Looks good to me.

PR #1866, opened 2023-01-29
Fix Japanese translation: ch-gates/basic-circuit-identities

kifumi:
Updated the wrong appearance of the equations on Japanese translation of `ch-gates/more-circuit-identities.ipynb`.


veenaiyuri (approved):
The original text includes "IBM Q devices" which needs to be corrected to "IBM Quantum devices" but as long as the translations are concerned, these look good to me.

kifumi:
@veenaiyuri Thank you for reviewing. I updated to "IBM Quantum" from "IBM Q".

veenaiyuri (approved):
LGTM

PR #1867, opened 2023-01-29
Fix Japanese translation: ch-gates/proving-universality

kifumi:
Updated the wrong appearance of the equations on Japanese translation of "ch-gates/proving-universality.ipynb."


veenaiyuri (approved):
Looks good to me.

0sophy1 (approved):
Thank you so much @kifumi !

PR #1868, opened 2023-01-29
Fix Japanese translation: ch-algorithms/quantum-phase-estimation

kifumi:
Updated the wrong appearance of the equations on Japanese translation of "ch-algorithms/quantum-phase-estimation.ipynb."


veenaiyuri (approved):
Looks good to me.

PR #1869, opened 2023-01-29
Fix Japanese translation: ch-gates/classical-computation-on-a-quantum-computer

kifumi:
Updated the wrong appearance of the equations on Japanese translation of "ch-gates/oracles.ipynb."


veenaiyuri (approved):
LGTM

PR #1880, opened 2023-02-01
Fix Japanese translation: /basics/single-systems.ipynb

kifumi:
Updated the wrong appearance of the equations on Japanese translation of basics/single-systems.ipynb.


veenaiyuri (approved):
Looks good to me

PR #1885, opened 2023-02-02
Add missing algorithm pages

### LINKED ISSUE #1673: Bug: two pages missing

qiskit.org/learn

Two notebooks are missing from the mega menu:
- [hidden-shift-problem.ipynb](https://github.com/Qiskit/platypus/blob/main/notebooks/ch-algorithms/hidden-shift-problem.ipynb)
- [quantum-walk-search-algorithm.ipynb](https://github.com/Qiskit/platypus/blob/main/notebooks/ch-algorithms/quantum-walk-search-algorithm.ipynb) (also missing from TOC)

### Screenshots

_No response_

### Additional information

_No response_

micpap25:
Fixes #1673

Adds the missing pages to the mega menu and table of contents.
For some reason the Quantum Walk Search Algorithm page isn't rendering on my machine... can someone check if this is something that needs to be fixed externally?


vabarbosa:
@micpap25 thank you!

@frankharkins is the order content/pages as you need them?

there is an unrelated issue with the `Quantum Walk` notebook which is causing it to not render properly. that could/should be tracked with a separate issue and PR. thanks!

frankharkins:
Sorry I missed this, @micpap25 is there a reason to put the hidden shift problem above Grover's algorithm?

micpap25:
I thought it made the most sense since the hidden shift problem directly builds on QFTs and so in terms of delivering content it would be better to cover it before Grover's algorithm.

frankharkins (approved):
Ok I see. LGTM, thanks!

PR #1887, opened 2023-02-03
Fix Japanese translation: /basics/single-systems.ipynb

kifumi:
Updated the wrong appearance of the equations on Japanese translation of basics/single-systems.ipynb.


kifumi:
@0sophy1 Thank you so much!

PR #1901, opened 2023-02-09
Fix Japanese translation: "intro" folder

kifumi:
- Updated the wrong appearance of the equations on Japanese translation of the files under `/translations/ja/intro` folder.
- Added missing folders of files, such as `images`, under `/translations/ja/intro` folder.
- Finished to check all the appearance is good in my local build environment.


veenaiyuri (approved):
LGTM

PR #1902, opened 2023-02-09
Add versioning banner

frankharkins:
## Changes

Display a banner if a course has more than one version.

[Preview link](https://platypus-pr-1902.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/course/ch-states)


frankharkins:
Feedback: Should use warning icon & grey banner instead of purple banner for old pages

JRussellHuffman:
My preference would be something like this. I think it's attention-grabbing enough, and it uses the standard component. 
Screen Shot 2023-02-15 at 4 11 05 PM

JRussellHuffman (changes requested):
I think have the banner be gray with a warning icon is the right approach, but open to discussion if it's too subtle. 

Screen Shot 2023-02-15 at 4 14 03 PM

JRussellHuffman (approved):
Everything for the scope of this PR looks good and works well. However, if I open the "browse all content" menu and navigate to another page while on the old version of the textbook, I am taken to the latest version of the textbook (not the old version). If I navigate to another page within the same unit (ie. it's listed on the left hand menu) while on the old version, it sticks with the old version. I think this is the prefered menu for the "browse all content" menu as well. I'm guessing that should be handled in a separate PR.

PR #1904, opened 2023-02-10
Fix Japanese translation: "ch-states" folder

kifumi:
- Updated the wrong appearance of the equations on Japanese translation of the files under `/translations/ja/ch-states` folder.


veenaiyuri (approved):
LGTM

PR #1909, opened 2023-02-11
Corrected one of the bell state bases

husayngokal:
I believe the last bell state was incorrect, as shown in the screenshot below:
Screen Shot 2023-02-11 at 22 23 48

I corrected it with the correct bell state, replacing |10> - |01> with |01> - |10>


husayngokal:
@vabarbosa I see that this hasn't been merged. Are there any issues?

PR #1910, opened 2023-02-12
Fir apostrophe typo

chutommy:
## Changes

I believe there should be

>the simulato**r’s** `.run()` method

instead of

>the simulato**rs** `.run()` method

## Screenshot

Untitled


frankharkins (approved):
Thanks

PR #1914, opened 2023-02-14
Fix custom problem set: Qiskit Beginner Proficiency

quantum-kittens:
## Changes

- Fixed text in Overview
- Fixed code in Problem 1


vabarbosa (approved):
thanks!

PR #1959, opened 2023-03-01
Add 'Entanglement in action' notebook

frankharkins:
## Changes

- Adds the new _Entanglement in action_ notebook
- Adjusts styling of images in blocks (see below)

Pending https://github.com/Qiskit/platypus/pull/1962

## Implementation details

Made a minor tweak to the css to allow the images in the tabs block to be larger.

Before / after

PR #1962, opened 2023-03-01
Update qiskit-terra req.

frankharkins:
## Changes

Updates `qiskit-terra` to `0.23.2` (required for #1959).

## Additional details

Unfortunately we can't use the `qiskit` meta package due to dependency conflicts with `qiskit-ibmq-provider` and the websockets package.


vabarbosa (approved):
thank you!

PR #1965, opened 2023-03-02
Variational Algorithm Design

pandasa123:
## Changes

Adding a few notes for us to keep in mind:

- LaTeX comments cause MathJAX issues
- Cells that require service authentication will fail in the CI. Add the `uses-hardware` tag to those cells

## How to read this PR

https://platypus-pr-1965.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/course/algorithm-design

### DIFF
--- a/notebooks/algorithm-design/instances/instances.ipynb
+++ b/notebooks/algorithm-design/instances/instances.ipynb
@@ -106,7 +106,7 @@
     "\n",
     "SSVQE's can be summarized as follows:\n",
     "\n",
-    "- Prepare several reference operators $U_R$ by apply one $U_R$ to several states to get many different $U_{R, j}$\n",
+    "- Prepare several reference states by applying a unitary U_R to k different computational basis states\n",
     "   - This algorithm requires the usage of $k$ mutually orthogonal reference states `{latex} \\{ |\\rho_j\\rangle \\}_{j=0}^{k-1}`, such that $\\langle \\rho_j | \\rho_l \\rangle = \\delta_{jl}$ for $j,l<k$.\n",
     "- Apply the variational form $U_V(\\vec\\theta_{i,j})$ to each reference state, resulting in the following ansatz $U_A(\\vec\\theta_{i,j})$.\n",
     "- Bootstrap at $i=0$ if a similar problem is available (usually found via classical simulation or sampling).\n",


notebooks/algorithm-design/applications.ipynb
```diff
@@ -0,0 +1,1534 @@
...
+   "metadata": {},
+   "source": []
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "## I'm following this how-to guide https://qiskit.org/documentation/partners/qiskit_ibm_runtime/how_to/noisy_simulators.html\n",
```
frankharkins, line 843: Cells from here onwards will fail during CI as we can't run them automatically. You can add a `uses-hardware` tag to tell our scripts to skip them.

frankharkins (commented):
I know this is still draft, but thought I'd leave some comments as I've seen some of these errors before. Feel free to ignore

notebooks/algorithm-design/concepts.ipynb
```diff
@@ -0,0 +1,544 @@
...
+    "\\hat{P}_k := \n",
+    "\\sigma_{k_{n-1}}\\otimes \\cdots \\otimes \\sigma_{k_0} \\quad \n",
+    "\\forall k \\in \\mathbb{Z}_4^n \\equiv \\{0,1,\\cdots,4^n-1\\}, \\\\\n",
+    "$$ \n",
+    "\n",
+    "such that $$\\hat{H} = \\sum^{4^n-1}_{k=0} w_k \\hat{P}_k,$$\n",
+    "\n",
+    "where $n$ is the number of qubits, $k \\equiv k_{n-1} \\cdots k_0$ for $k_l \\in \\mathbb{Z}_4 \\equiv \\{0, 1, 2, 3\\}$ (i.e. integers base $4$), and $(\\sigma_0, \\sigma_1, \\sigma_2, \\sigma_3) := (I, X, Y, Z)$.\n",
+    "\n",
+    "After performing this decomposition, `Estimator` derives a new circuit $V_k|\\psi\\rangle$ for each observable $\\hat{P}_k$ (i.e. from the original circuit), to effectively _diagonalize_ the Pauli observable in the computational basis and measure it. We can easily measure Pauli observables because we know $V_k$ ahead of time, which is not the case generally for other observables.\n",
```
jacobwatkins1, line 150: The basic principle is that observable O (I think O should be used) can usually be written as a sum of simpler things: O = A + B. And the crucial point is that EV(H) = EV(A) + EV(B). So we can compute EV(H) by dividing an conquering.

I think this point should be emphasized very early on. The fact that we decompose along Paulis is more of a detail.

notebooks/algorithm-design/concepts.ipynb
```diff
@@ -0,0 +1,544 @@
...
+    "\n",
+    "$$\n",
+    "\\hat{P}_k := \n",
+    "\\sigma_{k_{n-1}}\\otimes \\cdots \\otimes \\sigma_{k_0} \\quad \n",
+    "\\forall k \\in \\mathbb{Z}_4^n \\equiv \\{0,1,\\cdots,4^n-1\\}, \\\\\n",
+    "$$ \n",
+    "\n",
+    "such that $$\\hat{H} = \\sum^{4^n-1}_{k=0} w_k \\hat{P}_k,$$\n",
+    "\n",
+    "where $n$ is the number of qubits, $k \\equiv k_{n-1} \\cdots k_0$ for $k_l \\in \\mathbb{Z}_4 \\equiv \\{0, 1, 2, 3\\}$ (i.e. integers base $4$), and $(\\sigma_0, \\sigma_1, \\sigma_2, \\sigma_3) := (I, X, Y, Z)$.\n",
```
jacobwatkins1, line 148: I think this is too formal. Can we whittle down the notation a bit? Less tensor product symbols and set notation. Something like "with n qubits, I can take a string of n pauli operators". You can say there is an implicit tensor product.

notebooks/algorithm-design/concepts.ipynb
```diff
@@ -0,0 +1,544 @@
...
+    "for (circuit, pauli) in zip(aux_circuits, H.paulis):\n",
+    "    print(f\"Auxiliary circuit for {str(pauli)}\")\n",
+    "    display(circuit.draw(\"mpl\"))"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
```
jacobwatkins1, line 296: Can this code be turned interactive? Can we this code up into steps and have the reader either a) add code or b) fix incorrect code to do this themselves? Can we break this cell into multiple small cells for each task? (import statements, initialize state, create measurement circuits...)

Maybe call them "measurement circuits" instead of "auxiliary". Can each chunk of code be commented? 

Basically, do you hope that the reader will know how to do these sorts of tasks after reading this section? If so, more scaffolding will likely be required.

notebooks/algorithm-design/concepts.ipynb
```diff
@@ -0,0 +1,544 @@
...
+    "        print(f\"  >> Expected value of {str(obs)}: {expval:.5f}\")\n",
+    "    else:\n",
+    "        print(f\"  >> Total expected value: {expval:.5f}\")\n"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
```
jacobwatkins1, line 370: Similar with this block of code. Maybe Sampler and Estimator should be two separate tasks. An even more stripped down example for each might be appropriate.

notebooks/algorithm-design/concepts.ipynb
```diff
@@ -0,0 +1,544 @@
...
+    "### Mathematical rigor (optional)\n",
```
jacobwatkins1, line 371: This entire section is probably unnecessary. Except maybe this: the expectation value is simply the weighted average of the values of the observable with respect to the probability of measuring the value. This is an incredibly intuitive result and maybe deserves a proof. 

The second fact that might be worth "proving" is that expectation values are linear. But this follows directly from properties of matrix multiplication (or the inner product more abstractly). 

I would recommend something like in the attached photo:
image

jacobwatkins1 (commented):
Great start! My general feedback is the following:

- Focus more on circuits and expectation values in your "background" sections. These are necessary to understand your Runtime material. Rely on the UQuIC foundation more. 
- Make code section interactive. Turn it into a guided activity. Have reader fill in code themselves. Small additions by reader are still very valuable. Make problem easier as needed
- Reduce formality. Rely on intuitive concepts ("expectation values are the average value of the observable when measuring..."). Use only a little set theory notation.
- Mathematical Rigor section is almost entirely unnecessary. Only one or two theorems needed. State the theorem in clear, perhaps informal language, followed by proof, so that reader can get theorem without seeing proof. I've suggested a simplified proof in one of comments.

notebooks/algorithm-design/optimization.ipynb
```diff
@@ -0,0 +1,75 @@
...
+    "\n",
+    "Notice how we will only be able to minimize the cost function for the limited set of states that we are considering, if during our [choice of ansatz](ansatz.ipynb) we did not include the solution state in the set of searchable states, the values returned from the optimization routine will never correspond to the desired solution to our problem. But even if we included the ground state, we still need to be able to find it out of the set of searchable states —which is not a trivial task either."
```
frankharkins, line 14:
```suggestion
    "Notice how we will only be able to minimize the cost function for the limited set of states that we are considering. If our [choice of ansatz](ansatz.ipynb) can't create the solution state, the values returned from the optimization routine will never correspond to the solution to our problem. But even if the ansatz can create the ground state, we still need to find it, which is not a trivial task."
```

notebooks/algorithm-design/optimization.ipynb
```diff
@@ -0,0 +1,75 @@
...
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "\n",
+    "There are two main types of optimizers:\n",
+    "\n",
+    "* **Local**: these optimizers look for a point that minimizes the cost function starting at an initial point(s) $C(\\vec{\\theta_0})$ and moving to different points based on what they see in the region they happen to be at on successive iterations. That implies that the convergence of these algorithms will usually be fast, but can heavily dependent on the initial point. Some of these algorithms use the _gradient_ of the cost function $\\nabla C(\\vec{\\theta})$ (or an approximation) to choose the next set of values for the parameters $\\vec{\\theta}$ while others are based on completely different techniques like the _SIMPLEX_. Local optimizers are unable to see beyond the region were they are evaluating, and therefore turn out especially vulnerable to local minima; reporting convergence when they find one, ignoring other states with more favorable evaluations.\n",
```
frankharkins, line 38:
```suggestion
    "* **Local**: these optimizers look for a point that minimizes the cost function starting at an initial point(s) $C(\\vec{\\theta_0})$ and moving to different points based on what they see in the region they happen to be at on successive iterations. That implies that the convergence of these algorithms will usually be fast, but can heavily dependent on the initial point. Some of these algorithms use the _gradient_ of the cost function $\\nabla C(\\vec{\\theta})$ (or an approximation) to choose the next set of values for the parameters $\\vec{\\theta}$ while others are based on completely different techniques. Local optimizers are unable to see beyond the region were they are evaluating, and turn out to be especially vulnerable to local minima; reporting convergence when they find one, ignoring other states with more favorable evaluations.\n",
```

notebooks/algorithm-design/optimization.ipynb
```diff
@@ -0,0 +1,75 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Optimization loop\n",
```
frankharkins, line 8: I think this topic lends itself to visual descriptions, there are some good examples in the [QML course](https://learn.qiskit.org/course/machine-learning/training-quantum-circuits).

I also think this should be plural
```suggestion
    "## Optimization loops\n",
```

notebooks/algorithm-design/applications.ipynb
```diff
@@ -0,0 +1,1720 @@
...
+    "* `aux_operators_evaluated`: the expected values of any auxiliar operators we might include.\n",
+    "* `cost_function_evals`: Number of cost function evaluations.\n",
+    "* `eigenvalue`: the resulting eigenvalue.\n",
+    "* `optimal_circuit`: the parametrized circuit from which the solution was found. To find the eigenstates you have to bind the optimal values to the parameters.\n",
+    "* `optimal_parameters`: The set of optimal parameters as a dictionary\n",
+    "* `optimal_point`: an array with the optimal parameter values\n",
+    "* `optimal_value`: the optimal value of the cost function.\n",
+    "* `optimizer_evals`: number of optimizer evaluations.\n",
+    "* `optimizer_result`: results of the optimizer as an [`OptimizerResult`](https://qiskit.org/documentation/stubs/qiskit.algorithms.optimizers.OptimizerResult.html) object.\n",
+    "* `optimizer_time`: time used for optimization (in seconds)."
```
frankharkins, line 232: I'd just mention a couple of useful ones here (probably just `optimizer_result` and `optimal_point`) and link out to the API docs.

notebooks/algorithm-design/concepts.ipynb
```diff
@@ -0,0 +1,490 @@
...
+   "source": [
+    "This course is meant to be an introduction to quantum variational algorithms. Before we dive into the specifics of these algorithms, there are some quantum computation concepts that are essential to understand how and why they work. Most of them are covered in [Understanding Quantum Information and Computation](https://qiskit.org/learn/course/basics-quantum-information/) (also available as [a series of YouTube videos](https://www.youtube.com/playlist?list=PLOFEBzvs-VvqKKMXX4vbi4EB1uaErFMSO)). We strongly recommend that you go through the contents of that course first, as we will assume some familiarity with the theoretical context it provides. However, in this chapter we offer a quick reminder of some of the most important concepts. After that, we introduce Qiskit's *primitives*, the building blocks of variational and many other quantum algorithms."
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### General quantum systems\n",
```
frankharkins, line 24: You could consider trimming each of these headings down and putting them in a tabs block (such as the one at the top of [_What is quantum?_](https://learn.qiskit.org/course/introduction/what-is-quantum)).

frankharkins (commented):
Great work! Hopefully you find these comments helpful. You could consider adding annotations; some words I thought would be good candidates are
  - unitary
  - Hermitian
  - eigenstate
  - primitive
  - cost function
  - Pauli operator
  - spectral decomposition

\+ any others, I'll keep an eye out.

notebooks/algorithm-design/instances.ipynb
```diff
@@ -0,0 +1,284 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Subspace Search VQE (SSVQE)\n",
+    "\n",
+    "[SSVQE](https://arxiv.org/abs/1810.09434) is a variant of VQE that, like VQD, enables you to obtain the first $k$ eigenvalues of an observable $\\hat{H}$ with eigenvalues $\\{\\lambda_0, \\lambda_1,...,\\lambda_{N-1}\\}$, with $N\\geq k$. We'll assume without loss of generality that $\\lambda_0<\\lambda_1<...<\\lambda_{N-1}$.\n",
+    "\n",
+    "This algorithm requires the usage of $k$ mutually orthogonal reference states ${|\\rho_j\\rangle}_{j=0}^{k-1}$, that is, $\\langle \\rho_j | \\rho_l \\rangle = \\delta_{jl}$ for $j,l<k$. \n",
```
frankharkins, line 174:
```suggestion
    "This algorithm requires the usage of $k$ mutually orthogonal reference states $|\\rho_j\\rangle_{j=0}^{k-1}$, that is, $\\langle \\rho_j | \\rho_l \\rangle = \\delta_{jl}$ for $j,l<k$. \n",
```

notebooks/algorithm-design/instances.ipynb
```diff
@@ -0,0 +1,284 @@
...
+    "\n",
+    "The question that arises now is: How can we ensure that those states are all eigenstates? That's where the weights $w_j$ become important. The idea is that, if the weights are different enough, the term with largest weight (i.e. $w_0$) will be given priority during optimization over the rest, so the resulting state $U(\\vec{\\theta})|\\rho_0 \\rangle$ will end up becoming the eigenstate corresponding to $\\lambda_0$. Then, because of the mutual orthogonality of ${U(\\vec{\\theta})|\\rho_j\\rangle}_{j=0}^{k-1}$, the rest of the states will be orthogonal to it and, therefore, contained in the subspace corresponding to the eigenvalues $\\{\\lambda_1,...,\\lambda_{N-1}\\}$.\n",
```
frankharkins, line 192:
```suggestion
    "The question that arises now is: How can we ensure that those states are all eigenstates? That's where the weights $w_j$ become important. The idea is that, if the weights are different enough, the term with largest weight (i.e. $w_0$) will be given priority during optimization over the rest, so the resulting state $U(\\vec{\\theta})|\\rho_0 \\rangle$ will end up becoming the eigenstate corresponding to $\\lambda_0$. Then, because of the mutual orthogonality of $U(\\vec{\\theta})|\\rho_j\\rangle_{j=0}^{k-1}$, the rest of the states will be orthogonal to it and, therefore, contained in the subspace corresponding to the eigenvalues $\\{\\lambda_1,...,\\lambda_{N-1}\\}$.\n",
```

frankharkins (commented):
I think this will fix the errors on the instances page

notebooks/algorithm-design/ansatz.ipynb
```diff
@@ -0,0 +1,192 @@
...
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Ansatze and variational forms\n",
+    "\n",
+    "At the heart of all [variational algorithms](variational.ipynb) is the key idea of analyzing the differences between states; conveniently related through some well behaved mapping (e.g. continuous, differentiable) from a set of parameters or _variables_ —hence the name.\n",
+    "\n",
+    "In quantum computing, these algorithms work by exploring and comparing a range of _quantum states_ $|\\psi(\\vec{\\theta})\\rangle$, depending on a (finite) set of $k$ parameters $\\vec{\\theta} = (\\theta^0,...,\\theta^{k-1})$. These states can in turn be prepared through a _parametrized_ quantum circuit, representing certain parametric unitary transformation $U_V(\\vec{\\theta})$ from the [reference state](reference.ipynb), formally referred to as a _variational form_:\n",
+    "\n",
```
pandasa123, line 13: This is a comment

vandieren:
Paragraph 1.  Rephrase the eigenstate sentence so that it doesn’t involve 2 things in parenthesis and only one thing defined per sentence for better readability.  Suggested replacement: 
An eigenvector ( \(|\lambda\rangle\) ) of an observable is called an eigenstate.  Measuring an observable for one of its eigenstates ( \(|\lambda\rangle\)) will give us the corresponding eigenvalue (\(\lambda\)) as readout.

https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/concepts.ipynb?short_path=72bc6c0#L48

vandieren:
I feel the Basic Concepts section could be reworked, at least the part before Sampler and Estimator are introduced.  My suggestion is to take a new stab at it by reading through the course and figuring out what notation is used, and then limiting that section to those specifics.  We're already assuming knowledge of quantum, so starting with the reminder of an observable, eigenstates, eigenvalues is fine, but I'm not sure we need entanglement and superposition there.  Since the estimtator section talks about Pauli operators, adding a short paragraph with link about Pauli operators would be good to include in the Basic Concepts.

vandieren:
I recommend adding an introduction section before Basic Concepts, or at the start of the basic concepts section.  What is the course about?  The one sentence "This course introduces quantum variational algorithms." doesn't tell me why I should be excited to read on before I have to jump into review material.

vandieren:
typo: algorithm should be singular
"One of the main advantages of this kind of algorithms is..." 
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/variational.ipynb?short_path=13a9b76#L24

vandieren:
A generic circuit diagram might help illustrate the simplified hybrid workflow
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/variational.ipynb?short_path=13a9b76#L32

vandieren:
Typo missing "a" between "of" and "certain"
"A common goal of variational algorithms is to find the quantum state with the lowest or highest eigenvalue of certain observable"
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/variational.ipynb?short_path=13a9b76#L69

vandieren:
Either include information about spectral decomposition in the basic concepts section or link out to a source.  An undergrad who had only one semester of linear algebra (depending on the version of linear algebra they took) may not have encountered this concept.
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/variational.ipynb?short_path=13a9b76#L78

vandieren:
math <em> not rendering two times in the line that begins with:
"In the last step we used that,..."

https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/variational.ipynb?short_path=13a9b76#L80

vandieren:
semi-colon should be a comma "quantum circuit; such that"

https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/reference.ipynb?short_path=7e47e0a#L12

vandieren:
The phrase "For instance" doesn't introduce the section well.  I suggest just deleting that phrase.

https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/reference.ipynb?short_path=7e47e0a#L32

vandieren:
A sentence or two explaining what kinds of reference states are going to be introduced in this section and why one would chose one version over another could help prepare the reader for their journey through the list of examples.
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/reference.ipynb?short_path=7e47e0a#L12

vandieren:
semicolon should be a comma between "entanglement" and "like"
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/reference.ipynb?short_path=7e47e0a#L60

vandieren:
The typesetting for the Standard reference states is different than the rest of the page.
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/reference.ipynb?short_path=7e47e0a#L136

vandieren:
Since we talk about search space in the Heuristic section, I think it would help to add the role of the search space at the beginning of the page when we talk about analyzing the differences between states (ie better contextualize which states and what differences)
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/ansatz.ipynb?short_path=0f2fb59#L10

vandieren:
A few edits on the sentence beginning with "For a rotation layer"
"For a rotation layer, the gates are stacked on top of each other, while for an entanglement layer, the gate distribution can be selected by the user from several built-in distributions available in Qiskit."
Perhaps also, add a link to learn more about what those built-in distributions are.

https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/ansatz.ipynb?short_path=0f2fb59#L56

vandieren:
Edits: added commas and broke up a sentence.  
"Both types of layers can be parameterized or not, but at least one of them needs to contain parameters. Otherwise, without at least one parameter, there wouldn't be any variations!

Optionally, an extra rotation layer is added to the end of the circuit.

For example, let's create a $5$-qubit [`NLocal`](https://qiskit.org/documentation/stubs/qiskit.circuit.library.NLocal.html) circuit with rotation blocks formed by [`RX`](https://qiskit.org/documentation/stubs/qiskit.circuit.library.RXGate.html) and [`CRZ`](https://qiskit.org/documentation/stubs/qiskit.circuit.library.CRZGate.html) gates, entanglement blocks formed by [`Toffoli` gates](https://qiskit.org/documentation/stubs/qiskit.circuit.library.CCXGate.html) that act on qubits $[0,1,2]$, $[0,2,3]$, $[4,2,1]$ and $[3,1,0]$, and $2$ repetitions of each layer."

   
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/ansatz.ipynb?short_path=0f2fb59#L56-L90

vandieren:
edits: added comma and missing word
In order to evaluate said expectation value, we can then use the [Estimator primitive](https://github.com/qiskit-community/prototype-zne/blob/main/docs/tutorials/0-estimator.ipynb) in any of its flavors —albeit with some [tradeoffs](https://qiskit.org/documentation/partners/qiskit_ibm_runtime/tutorials/Error-Suppression-and-Error-Mitigation.html)

https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/optimization.ipynb?short_path=7c7dd62#L12

vandieren:
Adding a diagram showing three different possible outcomes (solution not covered by choice of ansatz, solution covered by choice of ansatz, but algorithm run doesn't hit it, solution covered by ansatz and algorithm converges) could help visualize this paragraph:
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/optimization.ipynb?short_path=7c7dd62#L14

vandieren:
This paragraph seems out of place and was covered in a previous section.  It could be deleted.
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/optimization.ipynb?short_path=7c7dd62#L23

vandieren:
Edits: apostrophe and colon added:
It's layout is simple: we label iterations within the optimization loop with the...

https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/instances.ipynb?short_path=3cd9489#L31

vandieren:
Rephrase the start of the first bullet point since "it's" didn't have a clear reference.

- For the initial step (i=0), ...

https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/instances.ipynb?short_path=3cd9489#L33

vandieren:
Some math is not rendering here
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/instances.ipynb?short_path=3cd9489#L36

vandieren:
typo: "get get"
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/instances.ipynb?short_path=3cd9489#L209

vandieren:
Reading through the course sequentially, I get to the end of Instances and extensions and wonder "what's runtime got to do with this?"  Also so far it's been all theory (which I personally love, but maybe not everyone will have the patience for that) with no example.

Can we start to build up a VQE algorithm step by step in each of the previous sections (code reference state, then build on that to the ansatz, demo that for various parameter values and see how the changing parameter values change the output, then go to the opti loop and demo that.)?

I'm still trying to think of how runtime and sampler can be introduced /motivated earlier.

vandieren:
Edits: 
The first equality holds because the ansatz $U(\vec{\theta})$ is a quantum operator and is therefore unitary. The last equality hold because of the orthogonality of the reference states. The fact that orthogonality is preserved through unitary transformations is deeply related to the principle of conservation of information, as expressed in quantum information science.

https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/instances.ipynb?short_path=3cd9489#L190

vandieren:
missing word "in" between "Fourier series" and "the following way"
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/instances.ipynb?short_path=3cd9489#L213

vandieren:
parameter misspelled
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/instances.ipynb?short_path=3cd9489#L221

vandieren:
Quantum Sampling Regression and the sampling to get the Fourier coefficients is probably a new technique that could be illustrated with an example (if it's important to understand).
https://platypus-pr-1965.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/course/algorithm-design/instances#instances-8-0

vandieren:
"are" is misspelled as "ar"

https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/applications.ipynb?short_path=1f8bc58#L318

vandieren:
"it seems reasonable" needs elaboration.  A reader who doesn't see why this is reasonable will be discouraged.  Also I suggest rephrasing the first part of the sentence as it reads awkwardly.  

Now that we have seen how a circuit preparing the solution state looks like, it seems reasonable to use a Hadamard gate as a reference circuit, so that the full ansatz becomes:



https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/applications.ipynb?short_path=1f8bc58#L387

vandieren:
The "however" doesn't seem to connect with anything previously.  What are we contrasting?
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/applications.ipynb?short_path=1f8bc58#L478

vandieren:
the word _verify_ should be _satisfy_
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/applications.ipynb?short_path=1f8bc58#L804

vandieren:
missing "a" between "as" and "proposed"
https://github.com/Qiskit/platypus/blob/e6b60b447ab1ab32ce0d27764f8576a8449d0350/notebooks/algorithm-design/applications.ipynb?short_path=1f8bc58#L836

vandieren:
Basic concepts - Sampler.  I love this description of the Sampler.  For someone new, it would be good to walk through a very basic example (maybe in a pop-open box that someone more familiar can just scroll past quickly) with code and output.  You can then refer to this example again when you want to motivate estimator for dense probability distributions to make it more tangible what you mean by "dense".  This example can also be leveraged in the estimator section when you describe how the estimator runs through all the Pk's computing the p_kj's to get the expectation value.

vandieren:
Basic concepts - Estimator primitive.  
"However, calculating the expectation value of an observable is not always possible, as we often don't know its eigenbasis."  This idea is referred to several times in the section and the reader may not know why we wouldn't "know" the eigenbasis.  Is there a reference we can link out to or an example we can give?

vandieren:
Variational algorithms - simplified hybrid workflow diagram fonts need to be bigger for readability.

vandieren:
LaTeX boxes (for instance on the Variational algorithm page) render with a strange border on the right and bottom using Safari.  This is particularly jarring in dark mode.

vandieren:
Variational algorithm LaTeX box, spell out "for all k" instead of using \forall

vandieren:
Optimization loop page: last sentence is missing a period.

vandieren:
Simplified hybrid workflow diagram and optimization loop -the convergence (or when to stop the loop) isn't very clear from the diagrams.

vandieren:
Optimization Loop - Measurement strategy... first sentence is missing a word "are"
"As mentioned, we using a noisy quantum computer as a black-box oracle, ..."

vandieren:
Optimization loop - Local and Global optimizers - Local Optimizers
"Local optimizers are unable to see beyond the region were they are evaluating, a..."  the word "were" should be where.

vandieren:
Instances and extensions - VQE
The bullet list describing the step of the VQE should have parallel structure.  Some bullets are sentences others begin with an action verb.  My recommendation is action verb.
- Set our optimization loop's **intial** parameter vectors at (i=0i=0i=0), also known as a "bootstrap" or "warm start," if we have a similar problem (typically found via classical simulation or sampling)

Also note the bold face typo in the example bullet above.

vandieren:
Instances and extension - VQD
Hermitian needs capitalized:
" We are adding this constraint because the eigenstates of an observable (i.e. hermitian operator)..."

vandieren:
Instances and extensions - SSVQD
"The last equality hold because of the orthogonality of the reference "
hold should be holds

vandieren:
Examples and Applications 
first line, extra word  (see bold)
"Now that we have gone through an overview of **the** all the parts that ..."

vandieren:
Examples and Applications
bullet list beginning with :
Here the parameters are:
- fun : the cost function value at the final step.

Remove the periods after each bullet point.

vandieren:
The last 2 lessons end abruptly.  Is there a way to signpost the next section or to point the learner to the assessments as a way to end the page?

JRussellHuffman:
Hey All! I've been reading through the Course and have some notes I'd like to share. Not sure if there are any differences, but I've been reading it on Saiba, not Platypus.

----
On the [first page](https://learning.www-dev.quantum-computing.ibm.com/course/near-term-algorithm-design-723/variational-algorithms-213#variational-algorithms), there is a paragraph that reads 

> “This course applies concepts covered in Basics of Quantum Information and Computation (also available as a series of YouTube videos). Feel free to refer back to the series at any time.“

I think we should swap it for something that makes it sound less like this course has a prerequisite, but instead makes it sound like the courses simply serve different purposes. Something like

> “This course is meant to be a starting place for developers to work with quantum computing. If you are looking for theoretical and foundational knowledge around quantum computing in general, consider going here (link to UQIC)."

----
In the "optimize parameters" I found a few spelling errors: “have an pre-existing optimial solution” should be “a” pre-existing “optimal”. 

“We'll use parametrized circuits define a variational form to represent a collection of parametrized states for our variational algorithm to explore.”

"define" should probably be “defined”

(note, I didn't check the whole course for spelling, I just noticed these).

----

Is it Ansatze or Ansatz? I see both used and I'm not sure the difference. A definition for this term in general would be good. 

----
It would also be good to use the annotation feature throughout the course for some of the terminologies introduced.

----
Can the lesson's be numbered?

JRussellHuffman (commented):
Have you tested exporting and hosting graphics like this as SVG?
(edit: for Saiba)

vabarbosa (approved):
thank you!

PR #1969, opened 2023-03-02
Change register names

frankharkins:
## Changes

- Changes the register names so they're valid with `qiskit-terra==0.22`
- Added the version table at the bottom

## Context

The original plan had been to upgrade the Terra version before this was released, but we've hit some dependency problems. This PR should be reverted once these are resolved.


vabarbosa (approved):
thank you!

PR #1971, opened 2023-03-03
Fix Japanese translation: "basics/single-systems" and "ch-states/representing-qubit-states"

kifumi:
- Updated the Japanese translation of two files: "basics/single-systems" and "ch-states/representing-qubit-states".


0sophy1 (approved):
Thank you so much @kifumi san!

kifumi:
@0sophy1 thank you!

PR #1972, opened 2023-03-03
Update qiskit and qiskit_ibm_runtime

frankharkins:
## Changes

As of yesterday, the latest `qiskit_ibm_runtime` no longer has a dependency conflict with `qiskit==0.41.1`, so we can upgrade both. This removes the need for #1969.

Also see https://github.com/Qiskit/platypus-binder/pull/33

`nb_autorun` output:
Screenshot 2023-03-03 at 14 02 22


vabarbosa (approved):
thank you!

PR #1976, opened 2023-03-03
Watkins lesson 4 suggestions

jacobwatkins1:
## Changes
I have made some editorial suggestions to Lesson 4 (Entanglement in Action) text of Understanding Quantum Information and Computation. Note that this version of the notebook is old; only content changes are suggested here.

## Implementation details
Mainly content changes (wording, equations, etc.).


jacobwatkins1:
In addition to the changes of this PR, I have some general comments regarding the current layout of the Lesson 4 text.
- Currently, the "Preliminaries" section (which I label Section 0) is in a drop down box. I think this section should be unboxed, as in my version, since it provides crucial context for readers new to the subject. I wouldn't say it counts as "footnote"-type information, and provides meaning to the Alice-Bob paradigm for beginners to the subject.
- On the other hand, I think the carousel boxes are excellently used.

PR #1987, opened 2023-03-07
Fix math problems with mathigon

### LINKED ISSUE #1983: Bug: LaTeX encoding showing up in sidebar

https://learn.qiskit.org/course/basics/multiple-systems#multiple-31-0

unrendered text shows up in the left hand nav bar. Furthermore, the text seems to be out of place on the nav bar.

### Screenshots

Screenshot 2023-03-06 at 12 04 46 PM

### Additional information

_No response_

frankharkins:
## Changes

PR #1966 introduced some math errors; this PR tweaks the equations to resolve them.

- Fixes #1983
- fixes a rendering bug in the [multiple systems page](https://learn.qiskit.org/course/basics/multiple-systems#multiple-systems-quantum-operations) (under _Controlled-unitary operations_).

I'm not 100% sure why this fix works, my working hypothesis is its caused by lines only containing `=`.


korgan00:
We had a similar bug in Saiba. There are two syntaxes for headings.

The most common is the one that uses `#`: `# my heading with "#"`

For example:
> # my heading with `#`

Also, you can use `-` or `=` as "underline" and it will convert the text above to a heading:
```
My Heading
=========
```

You can use just one `=`.
```
My Heading with "="
=
```

For example:

> My Heading with `=`
> =

If the math detector doesn't "block" that content correctly, that content will be broken.
In this case, before the `$$` there is a line with some spaces. As you know, for markdown a paragraph is a piece of text with one empty line above and below it. The two spaces force the `$$` to be part of the previous paragraph and the math renderer probably doesn't recognize it as a math block, then the headings are rendered.

frankharkins:
@korgan00 thanks! I've removed the whitespace as well.

vabarbosa (approved):
thank you

PR #1992, opened 2023-03-07
update usesHardware logic

### LINKED ISSUE #1989: Bug: Can't run sample code in Intro because notebook doesn't get my API key

1. Go to ...https://learn.qiskit.org/course/introduction/why-quantum-computing
2. Scroll down to ...Where are we now
image
4. See error ...above

I've gone into IBM Quantum and obtained an API KEY - I've also added it to my Qiskit Account properly, and it indicates it's saved. 

However the code will still not run and I cannot get it to acknowledge my key. I have refreshed the page. 

I am using Google Chrome browser

See above - cannot run sample code

Key should be acknowledged and I should be able to run sample code

### Screenshots

image

### Device

_No response_

### OS

_No response_

### Browser

Chrome

### Additional information

_No response_

vabarbosa:
## Changes

update the `usesHardware` logic to allow run cell to be re-enabled

Fixes #1989 

## Implementation details

`usesHardware` flag only disables run cell if no token is  available

## How to read this PR

- review `usesHardware` logic

### DIFF
--- a/frontend/vue/components/CodeExercise/CodeExercise.vue
+++ b/frontend/vue/components/CodeExercise/CodeExercise.vue
@@ -163,8 +163,8 @@ export default defineComponent({
     codeChanged (code: string) {
       this.code = code
       const codeOutput: any = this.$refs.output
-      codeOutput.needsApiToken(this.code).then((isNeeded: boolean) => {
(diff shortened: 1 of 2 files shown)


frankharkins (approved):
Logic looks good, thanks!

PR #1996, opened 2023-03-10
Update TSP function return statement

Astri-Cornish:
## Changes
Updated return statement template in TSP classical solution function for participants to align with the updates to the graders to make it clear to the participants what they should be returning for their function.


Astri-Cornish:
Hi @vabarbosa the small front-end change for the fixes we're going on problem 2.1 - thank you!

PR #2015, opened 2023-03-22
Update 1 Algo Queen Problem Set ID in TOC

quantum-kittens:
## Changes

Changed id from algo_queen_2023 to algo_queen_2023_qchallenge_notepad so participants don't accidentally access the link before time. (Algo Queen participants have little to no risk in stumbling upon this public repo.)


vabarbosa (approved):
👍🏽

PR #2039, opened 2023-03-29
add link to localization guide

omarcostahamido:
- currently a user will not find the localization documentation just by reading the readme or following its links, including inside the contributions guide. A simple line here would make that connection much easier, and I think it is better to create a direct link rather than to create a new section at the end of `CONTRIBUTING.md`, since it is not about improving the textbook's content but just make it more accessible!

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ In order to test the application locally, you can visit a specific course page, 
 
 ## Content contributions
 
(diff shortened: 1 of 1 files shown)


vabarbosa (approved):
thank you @omarcostahamido

omarcostahamido:
thank you for approving @vabarbosa

PR #2049, opened 2023-04-05
Update goal names in utokyo problem set

frankharkins:
These should be unique


kifumi:
@frankharkins Thank you so much! You are right!

PR #2064, opened 2023-04-11
Update basics course to the new structure

abdonrd:
## Changes

Update basics course to the new structure.
Checking if it works well in Platypus & Saiba.


abdonrd:
Ouch! Doesn't work... Maybe @frankharkins you could help?

abdonrd:
@frankharkins will do it on Qiskit/textbook. Thanks!

PR #2070, opened 2023-04-14
Skip cloning Qiskit/textbook in tests

frankharkins:
## Changes

This step was added back when we were using Git submodules, which we don't any more.


vabarbosa (approved):
thank you!!

PR #2081, opened 2023-04-20
Variational Algorithm Design Updates

pandasa123:
## Changes

We had excellent feedback from @ElePT, @mrossinek, @Cryoris on elements of the course + code implementation details

## Implementation details

- Updates based on `qiskit-nature==0.6.0`, thanks to @mrossinek
- `Session` logic + adding examples for optimization, thanks to @ElePT
- Fixes for ansatz construction + `statevector`, thanks to @Cryoris
- Separately, @vabarbosa is [updating the nature installation](https://github.com/Qiskit/platypus-binder/pull/35/files)

## How to read this PR

https://platypus-pr-2081.ecy0akwlcpw.us-south.codeengine.appdomain.cloud/course/algorithm-design/variational


notebooks/algorithm-design/ansatz/ansatz.ipynb
mrossinek, line 1: The other thing that I had mentioned was that I do not think it is really necessary to run the full driver here.
Since you are really only accessing the `num_spatial_orbitals` (which is an `int`) and `num_particles` (which is a `tuple[int, int]`), I think it would look "less scary" if you simply say something along the lines of:

> Assume we have a problem with 2 spatial orbitals and 2 electrons (1 alpha spin, 1 beta spin). To construct an ansatz for this problem, we can do this as follows....

And then simply do:
```python
num_spatial_orbitals = 2
num_particles = (1, 1)
```
and maybe put a comment as well as link to `ElectronicStructureProblem.num_spatial_orbitals` and `num_particles`, respectively somewhere

mrossinek: The same holds for the other 2 cases where you use Qiskit Nature

ElePT (commented):
LGTM! Thanks for your fast reply to our feedback.

notebooks/algorithm-design/cost_function/cost_function.ipynb
frankharkins, line 1: This notebook fails on cell 3

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 33
     29     return values
     32 theta_list = (2 * np.pi * np.random.rand(1, 8)).tolist()
---> 33 cost_function_vqe(theta_list)

Cell In[3], line 25, in cost_function_vqe(theta)
     22 backend = service.backend("ibmq_qasm_simulator")
     24 # Use estimator to get the expected values corresponding to each ansatz
---> 25 estimator = Estimator(session=backend)
     26 job = estimator.run(ansatz, observable, theta)
     27 values = job.result().values

TypeError: __init__() got an unexpected keyword argument 'session'
```

frankharkins (changes requested):
It looks like at least some of this code doesn't work, but I also can't test "applications", "optimization", or "instances" as they take too long to run. Can you re-run these notebooks to check they work?

notebooks/algorithm-design/cost_function/cost_function.ipynb
ElePT: 🤔 what `qiskit-ibm-runtime` version is this being run on? I use this code on a daily basis and it should work fine.

pandasa123: @ElePT, it's due to an importing the primitive version in an earlier cell and not importing the runtime estimator in this cell

frankharkins: @ElePT I used `0.9.3`

PR #2103, opened 2023-05-07
Fix Japanese translation: files under intro, ch-algorithms, and ch-applications folder

kifumi:
## Changes
Updated six Japanese translation files to fix the wrong equations and to fix the small difference from English version.

@KaoriNamba  @derwind Thank you for the information!


0sophy1 (approved):
Thank you so much, Japanese localization team @kifumi !

PR #2167, opened 2023-07-18
Add algorithm fundamentals links to mega menu

frankharkins:
## Changes

Adds three missing links


vabarbosa (approved):
thank you! should these also get added to qiskit.org megamenu: https://github.com/Qiskit/qiskit.org/blob/main/components/Learn/LearnMegaMenu.vue

Issue #2158, opened 2023-07-13
Create a banner to announce the new Learning Platform


y4izus:
### Feature description

To announce the new Learning Platform we need to add a banner right below the top menu. It should be the same as [the one we have on qiskit.org](https://github.com/Qiskit/qiskit.org/pull/3411)

### Additional context

The current banner we have is purple, so we need to adapt it to this new style

Wait to merge until Learning Platform team says we can merge

Do not add the new learning platform link until it's launched

The more info blog is still being written, so we should add the link when we have it

techtolentino:
Adding the `On hold` label for now, until it resurfaces as a priority

mackinziblank:
Updating w/ new designs to announce the Learning platform! 

2 parts - large graphic on Textbook home page, and smaller banners on individual pages.

Home page graphic:
<img width="855" alt="image" src="https://github.com/Qiskit/platypus/assets/59657965/fcd640c7-848c-48b5-9ea9-1b36ce2a0f9a">


Smaller banners (same as what will appear in docs, as described in https://github.com/Qiskit/qiskit_sphinx_theme/pull/530): 
<img width="932" alt="image" src="https://github.com/Qiskit/platypus/assets/59657965/0a41311d-8e2f-4b35-bfa8-c31a89d1ffba">

PR #2180, opened 2023-08-11
feat: 1XP banner

eddybrando:
**🚨 Important: Do not merge yet! Only merge when scheduled and after replacing the links.**

---

## Changes

This PR adds the 1XP banners to the course pages.

Closes https://github.com/Qiskit/platypus/issues/2158

image

### DIFF
--- a/frontend/vue/components/OneXP/OneXPBanner.vue
+++ b/frontend/vue/components/OneXP/OneXPBanner.vue
@@ -9,7 +9,7 @@
       />
       <a
         class="one-xp-banner__cta"
-        href="https://example.com"
+        href="https://learning.quantum-computing.ibm.com"
         rel="noopener noreferrer"
         target="_blank"
         @click="onClick"
--- a/frontend/vue/components/OneXP/OneXPBanner.vue
+++ b/frontend/vue/components/OneXP/OneXPBanner.vue
@@ -14,7 +14,7 @@
         target="_blank"
         @click="onClick"
       >
-        <span class="one-xp-banner__cta__label" v-text="`Check it out!`" />
+        <span class="one-xp-banner__cta__label" v-text="`Check it out`" />
         <LaunchIcon class="one-xp-banner__cta__icon" />
       </a>
     </div>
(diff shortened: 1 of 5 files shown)


frontend/vue/components/OneXP/OneXPBanner.vue
```diff
@@ -0,0 +1,79 @@
...
+    flex-direction: column;
+    gap: $spacing-03;
+
+    @include mq($from: large) {
+      flex-direction: row;
+    }
+  }
+
+  &__copy {
+    font-weight: bold;
```
techtolentino, line 62: same comment as in another PR, but this seems to be spec'd out to be 600...

server/templates/textbook.pug
```diff
@@ -14,6 +14,8 @@ html(lang=course.locale dir=dir)
       q-the-menu
 
     main.body
+      q-one-xp-banner
+
```
vabarbosa, line 18: for courses with multiple versions, we should get clarification if this is how it should look like (multiple spaced out banners)

image

vabarbosa (changes requested):
thank you @eddybrando looks good. can you get confirmation from design on how to treat the below scenario?

vabarbosa (dismissed):
@eddybrando thank you!!! 
only thing is to remember to update the [URL](https://github.com/Qiskit/platypus/blob/04e5a30b6ece2975863c7548b751da2fcf7e3c5e/frontend/vue/components/OneXP/OneXPBanner.vue#L12) before merging :-)
thanks again

frontend/vue/components/OneXP/OneXPBanner.vue
eddybrando: Similarly,

> As discussed offline, these small design details are been discussed with the design team. It is known that our codebase won't reflect exactly the Figma design due to some structural differences, but that's OK for now.

techtolentino (approved):
LGTM 🚀 

Thanks @eddybrando

y4izus (approved):

PR #2188, opened 2023-08-24
feat: remove `noreferrer`

eddybrando:
When linking to external links, we want to preserve `noopener` but remove `noreferrer`.

## Changes

This PR removes the [`noreferrer`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer) keyword for the external link since we want to be able to use the `Referer` header.

Related to https://github.com/Qiskit/qiskit.org/issues/3514


abdonrd (approved):
Thanks! It works in the preview!

